PR #141, opened 2024-10-01
feat: add lessons component to course show

wielopolski:
## Jira issue(s)

[https://selleolabs.atlassian.net/jira/software/projects/LC/boards/3?selectedIssue=LC-232](url)

## Overview
This PR adds a custom component for assigning and deleting lessons to a course and adds a setup custom css file for deleting unnecessary elements (sub modal in archive modal).

### DIFF
--- a/apps/admin/src/config/app.ts
+++ b/apps/admin/src/config/app.ts
@@ -124,7 +124,6 @@ export class AdminApp {
         },
         {
           resource: this.db.getResource("courses"),
-          // componentLoader,
           options: {
             ...coursesConfigOptions,
           },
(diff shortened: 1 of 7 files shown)


apps/admin/src/config/app.ts
```diff
@@ -123,18 +124,11 @@ export class AdminApp {
         },
         {
           resource: this.db.getResource("courses"),
+          // componentLoader,
```
typeWolffo, line 127: I think we don't need that :D

typeWolffo (approved):

PR #151, opened 2024-10-01
Fix lc 247 send api request with no use effect

tomaszbilka:
## Jira issue(s)

https://selleolabs.atlassian.net/browse/LC-247

## Overview

- It sends answers on click with no useEffect(). 
- It adds default values for useStates for single and multiple answers


wielopolski (approved):
GJ 🍻

PR #163, opened 2024-10-04
feat: Split Courses view into Available Courses and Your Courses

piotr-pajak:


apps/web/app/modules/Dashboard/Courses/CourseCard.tsx
```diff
@@ -16,6 +17,7 @@ type CourseCardProps = Pick<
   | "enrolled"
 > & {
   href: string;
+  completedLessonCount?: number;
```
typeWolffo, line 20: Can we move this inside the Pick<> type? As far as I can see, the course type has this field.

typeWolffo (approved):

apps/web/app/modules/Dashboard/Courses/CourseCard.tsx
piotr-pajak: No, because the `availableCourses` object doesn't contain the `completedLessonCount` property.

PR #173, opened 2024-10-09
feat: Add currency

typeWolffo:

### DIFF
--- a/apps/admin/src/adminResourceOptions/coursesOptions.ts
+++ b/apps/admin/src/adminResourceOptions/coursesOptions.ts
@@ -61,12 +61,12 @@ export const coursesConfigOptions: ResourceOptions = {
     "category_id",
     "author_id",
     "price_in_cents",
+    "currency",
     "created_at",
     "updated_at",
     "status",
     "file",
     "lessons",
-    "currency",
   ],
   properties: {
     title: {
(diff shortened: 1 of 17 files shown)


apps/web/app/modules/Dashboard/Courses/CourseCardButton.tsx
```diff
@@ -3,12 +3,14 @@ import { Button } from "~/components/ui/button";
...
 
 type CourseCardButtonProps = {
+  currency: string;
   enrolled: boolean;
   isAdmin: boolean;
   priceInCents: number;
 };
 
 const CourseCardButton = ({
+  currency,
```
typeWolffo, line 13: We need to pass it from `CourseCard` - check tsc error

typeWolffo (commented):
looks good - just fix the error with props

apps/admin/src/adminResourceOptions/coursesOptions.ts
```diff
@@ -64,6 +66,7 @@ export const coursesConfigOptions: ResourceOptions = {
     "status",
     "file",
     "lessons",
+    "currency",
```
typeWolffo, line 69: I think it would be nice to move it under the price instead of keeping this at the end of the list

piotr-pajak (approved):

PR #178, opened 2024-10-15
fix: migrations

typeWolffo:


piotr-pajak (approved):
Based on what we agreed on Slack, I'm approving the changes.

PR #179, opened 2024-10-16
feat: new admin - part 1

typeWolffo:
## Overview

This PR introduces the first part of new admin in LMS core.

## Changes 
 - new module in `apps/web` dedicated only to the admin part of app
 - new endpoints added to the api needed to replace adminjs actions
 
 ## Notes
 - In some places it is not the best code - it is still a WIP under refactoring


wielopolski (approved):
Looks good 🥇 
We need to think about adding consts for string values.

PR #190, opened 2024-10-23
Feat/pp batch of improvements

piotr-pajak:
## Preview
Screenshot 2024-10-23 at 23 20 21

https://github.com/user-attachments/assets/5d308da8-da3b-4178-9ead-9540e56f0c10


apps/web/app/modules/Courses/Lesson/Lesson.page.tsx
```diff
@@ -85,11 +103,19 @@ export default function LessonPage() {
...
+            switch (data?.isSubmitted) {
+              case true:
+                clearQuizProgress.mutate({ lessonId });
+                break;
+              case false:
+                submitQuiz.mutate({ lessonId });
+                setIsOpen(true);
+                break;
+              default:
+                break;
```
typeWolffo, line 116: it could be extracted to function (maybe with ts-pattern?) - what do you think? 🤔

typeWolffo (approved):
solving the fill in the blanks does not update the summary item - but I assume its because we dont have backend for that?
image

wielopolski (approved):
GJ 🍻 
In next step We need to extract all question types to CONST.

PR #198, opened 2024-10-29
feat: richtext edior

typeWolffo:
## Jira issue(s)

https://selleolabs.atlassian.net/browse/LC-300

## Overview
- replaced textarea fields with rich text editor
- added Viewer component for displaying rich text

## Screenshots
image
image


apps/web/app/components/RichText/Editor.tsx
```diff
@@ -0,0 +1,58 @@
...
+import EditorToolbar from "./toolbar/EditorToolbar";
+import { cn } from "~/lib/utils";
+
+interface EditorProps {
+  content?: string;
+  onChange: (value: string) => void;
+  placeholder?: string;
+  id?: string;
+  className?: string;
+}
```
piotr-pajak, line 13: **question**: Can we use `types` instead of `interfaces` throughout the application? 
I think it would be more consistent

piotr-pajak (approved):
LGTM 🚀

PR #214, opened 2024-11-06
feat: implement Gravatar component for user avatars

typeWolffo:
## Jira issue(s)

https://selleolabs.atlassian.net/browse/LC-335

## Screenshots

image
image


wielopolski (approved):
LGTM 🍻 
What do you think about keeping the link to the avatar in local storage, e.g. for 1 day?

typeWolffo:
> LGTM 🍻 What do you think about keeping the link to the avatar in local storage, e.g. for 1 day?

As we discussed offline - we don't need that.

PR #225, opened 2024-11-08
fix: saving options for the question

typeWolffo:
## Jira issue
https://selleolabs.atlassian.net/browse/LC-343

## Overview
- Fixed a bug that caused question options to not save when creating/updating.


wielopolski (approved):
LG 🥇

PR #231, opened 2024-11-13
fix: LC-345 create without image

typeWolffo:
## Jira issue(s)

https://selleolabs.atlassian.net/browse/LC-345

## Overview
- added possibility to create a course without a thumbnail
- added possibility to create a lesson without a thumbnail
- added placeholder image for course/lesson cards


wielopolski (approved):
LGTM 🍻

PR #234, opened 2024-11-18
feat: Student Statistics Dashboard

piotr-pajak:
## Jire issue
[LC-367](https://selleolabs.atlassian.net/browse/LC-367)

## Preview
Screenshot 2024-11-21 at 17 27 11

[LC-367]: https://selleolabs.atlassian.net/browse/LC-367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/api/src/statistics/statistics.service.ts
```diff
@@ -1,12 +1,162 @@
...
+    const lastActivityDate = currentStats?.lastActivityDate
+      ? startOfDay(new Date(currentStats.lastActivityDate))
+      : null;
+
+    const newCurrentStreak = (() => {
+      if (!lastActivityDate) return 1;
+
+      const daysDiff = differenceInDays(today, lastActivityDate);
+
+      if (daysDiff === 0) return currentStats?.currentStreak ?? 1;
```
mateuszszczecina, line 84: Maybe we should use switch case here, it will be clearer and will work better if there are more conditions in the future

`switch (daysDiff) {
  case 0:
    return currentStats?.currentStreak ?? 1;
  case 1:
    return (currentStats?.currentStreak ?? 0) + 1;
  default:
    return 1;
}
`

apps/web/app/api/generated-api.ts
```diff
@@ -1159,6 +1171,24 @@ export interface GetUserStatisticsResponse {
...
     };
+    lastLesson: {
+      /** @format uuid */
+      id: string;
+      title: string;
+      imageUrl?: string;
+      description: string;
+      itemsCount: number;
+      itemsCompletedCount?: number;
+      lessonProgress?: "completed" | "in_progress" | "not_started";
```
mateuszszczecina, line 1182: It would be good to create a type for this
`type LessonProgress = "completed" | "in_progress" | "not_started";`

apps/web/app/components/Gravatar.tsx
```diff
@@ -3,19 +3,21 @@ import CryptoJS from "crypto-js";
 import { AvatarImage } from "./ui/avatar";
 
 type GravatarProps = {
-  email: string;
+  email: string | undefined;
```
mateuszszczecina, line 6: email?: string

apps/web/app/modules/Statistics/components/ChartLegendBadge.tsx
```diff
@@ -0,0 +1,15 @@
+import type { ReactNode } from "react";
+
+type ChartLegendBadgeProps = {
+  label: ReactNode | string | undefined;
```
mateuszszczecina, line 4: label?: ReactNode | string
dotColor?: string

apps/web/app/modules/Statistics/components/RatesChart.tsx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+type RatesChartProps = {
+  isLoading?: boolean;
+  resourceName: string;
+  chartData: ChartData[];
+};
+
+export const RatesChart = ({ isLoading = false, resourceName, chartData }: RatesChartProps) => {
+  const dataMax = Math.max(...chartData.map((d) => d.started));
```
mateuszszczecina, line 33: Change the name to something more descriptive. d -> data. 

Also i think it is a good idea to use useMemo in these 4 variables

apps/web/app/modules/Statistics/components/RatesChart.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const ticks = Array.from({ length: Math.floor(yAxisMax / step) }, (_, i) => (i + 1) * step);
+
+  if (isLoading) {
+    return (
+      <div className="w-full bg-white rounded-lg gap-4 drop-shadow-card p-8 flex flex-col">
+        <hgroup className="gap-y-[5px] flex flex-col items-center py-3">
+          <Skeleton className="h-6 w-[240px] rounded-lg" />
+          <Skeleton className="w-40 h-4 rounded-lg" />
+        </hgroup>
+        {/*<div className=""></div>*/}
```
mateuszszczecina, line 45: remove unused code

apps/web/app/api/generated-api.ts
piotr-pajak: As the filename indicates, this is a generated API based on the backend schema, so I can't modify the type here 😉

apps/web/app/components/Gravatar.tsx
piotr-pajak: I prefer using `email: string | undefined` instead of `email?: string` because the email field should always be included in the object, even if its value is undefined. 
With `email?: string`, the property can be completely omitted, which doesn't meet the requirement of ensuring the field is always present. 
Using `email: string | undefined` makes the intent clear that the property is required but can have an `undefined` value.

apps/web/app/modules/Statistics/components/ChartLegendBadge.tsx
piotr-pajak: Same case as above 😉

apps/web/app/modules/Statistics/utils.ts
```diff
@@ -0,0 +1,13 @@
+type Data =
```
typeWolffo, line 1: we need to add `object` to the union just to satisfy typescript, which is not ideal solution but it will work until we figure out how to generate better types from typebox schema

mateuszszczecina (approved):

wielopolski (approved):

apps/web/app/modules/Statistics/components/ProfileWithCalendar.tsx
```diff
@@ -0,0 +1,88 @@
...
+          </div>
+          <span className="text-neutral-800 body-lg-md">Daily Streak</span>
+        </div>
+        <div className="flex justify-center w-full">
+          <Calendar
+            mode="single"
+            showOutsideDays
+            fixedWeeks
+            weekStartsOn={1}
+            dates={Object.keys(streak?.activityHistory || {}) ?? []}
```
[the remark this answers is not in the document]
typeWolffo: I think we can safely apply it in the new PR to avoid possible breaking bugs before the demo 😉

typeWolffo (approved):

PR #239, opened 2024-11-21
fix: enhance quiz answer checks in e2e tests to verify input state and visibility of missing answers

typeWolffo:
## Overview
- Fixed an issue in the quiz test where options were clicked too quickly without checking.


wielopolski (approved):
GJ 🍻

PR #308, opened 2024-12-20
fix: thumbnail display

mateuszszczecina:
## Jira issue(s)

https://selleolabs.atlassian.net/browse/LC-514)

## Overview

Fix problem with display thumbnail in course settings

### DIFF
--- a/apps/api/src/courses/schemas/showCourseCommon.schema.ts
+++ b/apps/api/src/courses/schemas/showCourseCommon.schema.ts
@@ -20,6 +20,8 @@ export const commonShowCourseSchema = Type.Object({
   isScorm: Type.Optional(Type.Boolean()),
   priceInCents: Type.Number(),
   thumbnailUrl: Type.Optional(Type.String()),
+  thumbnailS3Key: Type.Optional(Type.String()),
+  thumbnailS3SingedUrl: Type.Optional(Type.String()),
   title: Type.String(),
 });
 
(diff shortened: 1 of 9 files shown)


apps/api/src/courses/schemas/showCourseCommon.schema.ts
```diff
@@ -20,6 +20,8 @@ export const commonShowCourseSchema = Type.Object({
   isScorm: Type.Optional(Type.Boolean()),
   priceInCents: Type.Number(),
   thumbnailUrl: Type.Optional(Type.String()),
```
mateuszszczecina, line 22: For now I also left the old value because it is used in other places and I am not sure if it is needed for now, it will be possible to remove it when we completely switch to the new course backend

apps/web/app/modules/Admin/EditCourse/CourseSettings/hooks/useCourseSettingsForm.tsx
```diff
@@ -14,46 +13,35 @@ type CourseSettingsProps = {
...
-  const newImageExist = Boolean(newImageUrl);
-  const imageUrlChanged = oldImageExist && newImageExist && newImageUrl !== oldImageUrl;
-
   const onSubmit = async (data: UpdateCourseBody) => {
     updateCourse({
       data: { ...data },
       courseId,
     }).then(() => {
       queryClient.invalidateQueries(courseQueryOptions(courseId));
-      if (imageUrlChanged) {
```
mateuszszczecina, line 54: I removed it because it seems to me that it doesn't work at the moment

typeWolffo (approved):

piotr-pajak (approved):

PR #316, opened 2024-12-30
feat: 519 question match words

mateuszszczecina:
## Jira issue(s)
https://selleolabs.atlassian.net/browse/LC-519

## Overview
Add new quiz type -> match words


apps/api/src/lesson/repositories/adminLesson.repository.ts
```diff
@@ -170,6 +172,66 @@ export class AdminLessonRepository {
...
   }
 
+  async getExistingQuestions(lessonId: UUIDType) {
+    return await this.db
+      .select({ id: questions.id })
+      .from(questions)
+      .where(eq(questions.lessonId, lessonId));
+  }
+
+  async upsertOptions(questionId: UUIDType) {
```
wielopolski, line 182: Its not a upsert method

apps/api/src/lesson/repositories/adminLesson.repository.ts
```diff
@@ -170,6 +172,66 @@ export class AdminLessonRepository {
...
+    await this.db.delete(questions).where(inArray(questions.id, questionsToDelete));
+  }
+
+  async deleteQuestionOptions(questionsToDelete: UUIDType[]) {
+    await this.db
+      .delete(questionAnswerOptions)
+      .where(inArray(questionAnswerOptions.questionId, questionsToDelete));
+  }
+
+  async upsertQuestion(questionData: QuestionBody, lessonId: UUIDType, authorId: UUIDType, questionId?: UUIDType): Promise<UUIDType> {
```
wielopolski, line 221: Maybe we can use drizzle upsert to do it?
```suggestion
  await db.insert(users)
  .values({ id: 1, name: 'Dan' })
  .onConflictDoUpdate({ target: users.id, set: { name: 'John' } });
```

apps/api/src/lesson/services/adminLesson.service.ts
wielopolski, line 1: Great job with extracted DB methods to Repository, but I think that we need to use trx and provide it to call of method

wielopolski (changes requested):
GJ 🥇

wielopolski (approved):
LGTM 🍻

PR #374, opened 2025-01-14
fix: problem with dnd and order

mateuszszczecina:
## Overview

Task from docs, created by Tomasz Czana. 

When bottom question is opened (editing) and the rest are closed, when I drag and drop the bottom one and move it to the top, the new bottom question will expand and mine that I move to the top collapses.


wielopolski (changes requested):
@piotr-pajak has prepared a pretty good refactor compare proposed changes and wait for his changes on main branch.

piotr-pajak (approved):

wielopolski (approved):
LGTM 🍻

PR #376, opened 2025-01-14
fix: Drag and drop reordering logic

piotr-pajak:
## Jira issue(s)
[LC-572](https://selleolabs.atlassian.net/jira/software/projects/LC/boards/3?selectedIssue=LC-572)

## Overview:
- `SortableList.tsx`: Improved type safety by making id fields mandatory and replaced interfaces with type aliases.
- `QuizLessonForm.tsx`: Reordered declarations and added temporary id to new options using `crypto.randomUUID()`.
- `QuizLessonForm.types.ts`: Made id mandatory in QuestionOption and Question types.
- Components: Updated several components to use id instead of displayOrder and added id to new options using `crypto.randomUUID()`.

[LC-572]: https://selleolabs.atlassian.net/browse/LC-572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/api/src/chapter/repositories/adminChapter.repository.ts
+++ b/apps/api/src/chapter/repositories/adminChapter.repository.ts
@@ -67,11 +67,11 @@ export class AdminChapterRepository {
   async removeChapter(chapterId: UUIDType, trx?: PostgresJsDatabase<typeof schema>) {
     const dbInstance = trx ?? this.db;
 
-    return await dbInstance.delete(chapters).where(eq(chapters.id, chapterId)).returning();
+    return dbInstance.delete(chapters).where(eq(chapters.id, chapterId)).returning();
   }
 
   async getBetaChapterLessons(chapterId: UUIDType): Promise<AdminLessonWithContentSchema[]> {
-    return await this.db
+    return this.db
       .select({
         updatedAt: sql<string>`${lessons.updatedAt}`,
         id: lessons.id,
@@ -104,12 +104,14 @@ export class AdminChapterRepository {
                   )
                   FROM ${questionAnswerOptions} questionAnswerOptions
                   WHERE questionAnswerOptions.question_id = questions.id
+                  ORDER BY ${questionAnswerOptions.displayOrder}
                 )
               )
             )
             FROM ${questions}
             WHERE ${questions.lessonId} = lessons.id
-          )
+            ORDER BY ${questions.displayOrder}
+                 )
         )`,
       })
       .from(lessons)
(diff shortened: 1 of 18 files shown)


wielopolski (approved):
GJ 🍻 
Do I understand correctly that we can now change the Question type and require id?
```
export const adminQuestionSchema = Type.Object({
  id: Type.Optional(UUIDSchema),
  type: Type.Enum(QUESTION_TYPE),
  description: Type.Optional(Type.Union([Type.String(), Type.Null()])),
  title: Type.String(),
  displayOrder: Type.Optional(Type.Number()),
  solutionExplanation: Type.Optional(Type.String()),
  photoS3Key: Type.Optional(Type.Union([Type.String(), Type.Null()])),
  options: Type.Optional(Type.Array(adminOptionSchema)),
});
```

mateuszszczecina:
I checked it on your branch and adding new questions and options inside the course does not work. I am not sure if generating ids for objects that are not in the database is a good approach. In this approach we send the id to the backend and it treats it as an object that it already has and wants to edit it but we do not have it. Another problem is that these ids are not unique enough that even if we manage to save them in the database there is a small chance that such an id already exists in the database. In my opinion we should avoid this approach. My idea is to create a dummy id e.g. sortableId that will be generated on the frontend for questions and options and will not be sent to the backend, this will protect us from unwanted behavior.

mateuszszczecina (changes requested):
I checked it on your branch and adding new questions and options inside the course does not work, it does not create new options or new questions. 
I am not sure if generating ids for objects that are not in the database is a good approach. In this approach we send the id to the backend and it treats it as an object that it already has and wants to edit it but we do not have it. Another problem is that these ids are not unique enough that even if we manage to save them in the database there is a small chance that such an id already exists in the database. In my opinion we should avoid this approach. My idea is to create a dummy id e.g. sortableId that will be generated on the frontend for questions and options and will not be sent to the backend, this will protect us from unwanted behavior. @wielopolski @typeWolffo Please express your opinion

wielopolski:
Idea is quite alright, but I suggest using some kind of additional property that will not have an impact on data that is sent to db record

mateuszszczecina (approved):

apps/api/src/chapter/repositories/adminChapter.repository.ts
```diff
@@ -104,12 +104,14 @@ export class AdminChapterRepository {
...
                   WHERE questionAnswerOptions.question_id = questions.id
+                  ORDER BY ${questionAnswerOptions.displayOrder}
                 )
               )
             )
             FROM ${questions}
             WHERE ${questions.lessonId} = lessons.id
-          )
+            ORDER BY ${questions.displayOrder}
+                 )
```
mateuszszczecina, line 114: i think something goes wrong with formatting

PR #413, opened 2025-02-08
fix: update from mail address

wielopolski:
## Overview
This PR removes disallow email address in setup mailer.

[LC-223]: https://selleolabs.atlassian.net/browse/LC-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/app/api/api-client.ts
+++ b/apps/web/app/api/api-client.ts
@@ -20,6 +20,7 @@ ApiClient.instance.interceptors.request.use((config) => {
   const isAuthEndpoint =
     config.url?.includes("/login") ||
     config.url?.includes("/refresh") ||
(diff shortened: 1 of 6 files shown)


apps/api/src/courses/course.service.ts
```diff
@@ -360,7 +360,7 @@ export class CourseService {
   }
 
   async getCourse(id: UUIDType, userId: UUIDType): Promise<CommonShowCourse> {
-    //TODO: to remove
+    //TODO: to removegpf
```
PoslinskiNet, line 363: what this change does?

PR #425, opened 2025-03-06
chore_add_environment_variable_to_Sentry

MarcinMaciejewski337:
## Jira issue(s)
[LC-606]()

## Screenshots

[LC-606]: https://selleolabs.atlassian.net/browse/LC-606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/api/src/main.ts
+++ b/apps/api/src/main.ts
@@ -18,6 +18,7 @@ async function bootstrap() {
     integrations: [nodeProfilingIntegration()],
     tracesSampleRate: 1.0,
     profilesSampleRate: 1.0,
+    environment: process.env.NODE_ENV === "production" ? "production" : "development",
   });
 
   const app = await NestFactory.create(AppModule, {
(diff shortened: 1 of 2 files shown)


apps/api/src/main.ts
```diff
@@ -18,6 +18,7 @@ async function bootstrap() {
     integrations: [nodeProfilingIntegration()],
     tracesSampleRate: 1.0,
     profilesSampleRate: 1.0,
+    environment: process.env.NODE_ENV === "production" ? "production" : "development",
```
wielopolski, line 21: It seems to me that we should still add support for staging environment

PoslinskiNet (approved):

qbart:
would be better to rely on process.env.SENTRY_ENVIRONMENT as official documentation suggests

PR #427, opened 2025-03-07
fix: add max length validation for description field with user feedback in NewLesson form

typeWolffo:
## Overview
- added validation for lesson content to match with DB limits
[Sentry](https://selleolabs.sentry.io/issues/32061468/?project=4508880341106768)

### Before
image
### After
image


mateuszszczecina:
I think we should also implement validation on backend side

typeWolffo:
@mateuszszczecina there was only BE validation so thats why the error occurs

mateuszszczecina:
I mean we should return a more descriptive error from the backend and a different error code than 500

mateuszszczecina (approved):

PR #432, opened 2025-03-17
chore: increase file size upload limit

maciejhubert:
Increasing the limit of uploaded videos, as in real case videos will usually have more than 20MB


mikoscz:
FYI: the Multer defaults to memory storage: https://github.com/expressjs/multer/blob/master/index.js#L17 for bigger files I would recommend streaming them through to S3 or uploading them directly, with that limit set we can exhaust memory. Probably it would be good to address that in the future.

PR #436, opened 2025-03-28
chore: increase character limit in text editor to 5000

maciejhubert:
Currently character limit in text editor (text lesson) is not enough, especially in case of special characters such as checkbox which takes 150 characters (to be fixed another time).

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,15 +47,15 @@ A modern, scalable Learning Management System built with cutting-edge technologi
 
 </div>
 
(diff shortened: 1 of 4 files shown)


wielopolski (changes requested):
After the change in index.ts, execute `pnpm drizzle-kit generate --name=<name>`
and then `pnpm drizzle-kit migrate`.
This will allow the changes to be inserted correctly into the database.
In addition, it is actually worth changing the type to text and using the restrictions in the code.

PR #437, opened 2025-03-28
chore: update translation and description column in lessons

wielopolski:
Currently character limit in text editor (text lesson) is not enough, especially in case of special characters such as checkbox which takes 150 characters (to be fixed another time).


apps/web/app/modules/Admin/EditCourse/CourseLessons/NewLesson/TextLessonForm/TextLessonForm.tsx
```diff
@@ -46,7 +46,7 @@ const TextLessonForm = ({
     setIsModalOpen(true);
   };
 
-  const maxDescriptionFieldLength = 3000;
+  const maxDescriptionFieldLength = 5000;
```
piotr-pajak, line 49:
```suggestion
  const MAX_DESCRIPTION_FIELD_LENGTH = 5000;
```

piotr-pajak (approved):

PR #444, opened 2025-05-17
Mm fix lc 620 lesson is not refreshed when switching between lessons

MarcinMaciejewski337:
## Jira issue(s)
[LC-620](https://selleolabs.atlassian.net/browse/LC-620)

### DIFF
--- a/apps/web/app/components/RichText/Editor.tsx
+++ b/apps/web/app/components/RichText/Editor.tsx
@@ -1,4 +1,5 @@
 import { EditorContent, useEditor } from "@tiptap/react";
+import { useEffect } from "react";
 
 import { cn } from "~/lib/utils";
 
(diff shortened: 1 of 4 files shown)


typeWolffo (approved):

apps/web/app/components/RichText/Editor.tsx
```diff
@@ -28,6 +29,12 @@ const Editor = ({ content, placeholder, onChange, id, className }: EditorProps)
     },
   });
 
+  useEffect(() => {
+    if (editor && content !== editor.getHTML()) {
+      editor.commands.setContent(content || "");
+    }
+  }, [content]);
```
typeWolffo, line 36: React Hook useEffect has a missing dependency: 'editor'. Either include it or remove the dependency array

if intended you can disable this warning with note why

wielopolski:
In addition, we had a similar problem and could only add the container key to distinguish the change in content

maciejhubert:
Sorry guys, I rushed a bit with the merge as I saw a green check - @MarcinMaciejewski337 will apply your feedback in the next PR 🙏

PR #480, opened 2025-07-02
fix(SCS-472): Removed unnecessary sections in profile for student and fixed undefined in breadcrumbs

Japrolol:
## Jira issue(s)
[SCS-472]()
https://github.com/Selleo/mentingo/issues/472
## Overview
Added a helper function to frontend for role checking. Fixed backend, now it correctly returns the user details, even if you are a student. Added conditional checks to the profile page and finally fixed the teacher route to profile.

## Screenshots
Student View:
Screenshot 2025-07-02 at 13 11 49
Screenshot 2025-07-02 at 13 12 13

Teacher View:
Screenshot 2025-07-02 at 13 12 36
Screenshot 2025-07-02 at 13 12 51

There are some stuff that need to be clarified, I left comments, have already asked for feedback


apps/api/src/user/schemas/user.schema.ts
```diff
@@ -11,6 +11,7 @@ export const userDetailsSchema = Type.Object({
   contactEmail: Type.Union([Type.String(), Type.Null()]),
   contactPhone: Type.Union([Type.String(), Type.Null()]),
   jobTitle: Type.Union([Type.String(), Type.Null()]),
+  role: Type.Union([Type.String(), Type.Null()]),
```
piotr-pajak, line 14: You can replace `Type.String()` with `Type.Enum(USER_ROLES)` since we already have a defined schema for user roles 💡

BTW. I'm not sure if role should actually be nullable, it seems like a required field for using the app properly.
```suggestion
  role: Type.Union([Type.Enum(USER_ROLES), Type.Null()]),
```

apps/api/src/user/user.service.ts
```diff
@@ -95,23 +95,26 @@ export class UserService {
       .select({
         firstName: users.firstName,
         lastName: users.lastName,
-        id: userDetails.id,
+        role: users.role,
+        id: users.id,
```
piotr-pajak, line 99: Do we need to include id: users.id in userDetails if we're already getting the user ID from the current user query? Might be cleaner to remove it if it's redundant. 🤔 ✂️ 
```suggestion
```

apps/web/app/api/queries/useContentCreatorCourses.ts
```diff
@@ -9,15 +9,23 @@ type SearchParams = {
...
+export const ContentCreatorCoursesOptions = (
+  authorId?: string,
+  searchParams?: SearchParams,
+  isContentCreator?: boolean,
+) => {
   return {
-    enabled: !!authorId,
-    queryKey: ["content-creator-courses", authorId],
+    enabled: !!authorId && isContentCreator,
+    queryKey: ["contentCreator-courses", authorId],
```
piotr-pajak, line 19: Just a question, is this consistent with our existing queryKey naming? 
We're mixing camelCase and kebab-case here. 
If camelCase is the standard, then this change looks good!

apps/web/app/utils/userRoles.ts
piotr-pajak, line 1: We already have a `useUserRole` hook that provides this info, so duplicating the role helpers here might be unnecessary. 
Also, the comment about 'potentially shared between frontend and backend' doesn’t quite apply until we actually use it in both places, for now, it's just duplication. 
However, if we really needed cross-environment sharing, we would place this in @repo/shared.

apps/web/app/modules/Profile/Profile.page.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { Button } from "~/components/ui/button";
+import CourseCard from "~/modules/Dashboard/Courses/CourseCard";
+import { isAdminLike } from "~/utils/userRoles";
+
+import { ProfilePageBreadcrumbs } from "./ProfilePageBreadcrumbs";
+
+export default function ProfilePage() {
+  const { id = "" } = useParams();
+  const { data: userDetails } = useUserDetails(id);
+  const hasPermission = isAdminLike(userDetails?.role ?? "");
```
piotr-pajak, line 20:
```suggestion
  const { isAdminLike } = useUserRole()
  const hasPermission = isAdminLike
```
or even simplier:
```suggestion
  const { isAdminLike } = useUserRole()
```
(explanation below)

apps/web/app/modules/Profile/Profile.page.tsx
```diff
@@ -0,0 +1,121 @@
...
+      {userDetails && (
+        <ProfilePageBreadcrumbs
+          id={id}
+          username={`${userDetails?.firstName} ${userDetails?.lastName}`}
+        />
+      )}
+      <div className="flex flex-col gap-6 xl:flex-row">
+        <section className="flex flex-col gap-y-6 rounded-b-lg rounded-t-2xl bg-white p-6 drop-shadow xl:w-full xl:max-w-[480px]">
+          <div className="flex flex-col gap-6 md:flex-row md:items-center">
+            <Avatar className="h-20 w-20">
```
piotr-pajak, line 34:
```suggestion
            <Avatar className="size-20">
```

apps/web/app/modules/Profile/Profile.page.tsx
```diff
@@ -0,0 +1,121 @@
...
+                  </p>
+                )}
+              </div>
+            </div>
+          </div>
+          {hasPermission && (
+            <div className="flex flex-col gap-y-2">
+              <div className="flex items-center gap-x-3">
+                <span className="text-neutral-900">{t("contentCreatorView.other.about")}</span>
+                <div className="h-[1px] w-full bg-primary-200" />
```
piotr-pajak, line 54:
```suggestion
                <div className="h-px w-full bg-primary-200" />
```

apps/web/app/modules/Profile/Profile.page.tsx
```diff
@@ -0,0 +1,121 @@
...
+              </div>
+              <p className="body-base mt-2 text-neutral-950">{userDetails?.description}</p>
+            </div>
+          )}
+          {hasPermission && (
+            <div className="flex flex-col gap-y-1 md:gap-y-4 xl:mt-auto">
+              <div className="flex items-center gap-x-3">
+                <span className="text-neutral-900">{t("contentCreatorView.other.contact")}</span>
+                <div className="h-[1px] w-full bg-primary-200" />
```
piotr-pajak, line 63:
```suggestion
                <div className="h-px w-full bg-primary-200" />
```

apps/web/app/modules/Profile/Profile.page.tsx
```diff
@@ -0,0 +1,121 @@
...
+              </div>
+              <div className="flex flex-col gap-3 md:flex-row md:*:w-full">
+                <a
+                  href={`tel:${userDetails?.contactPhone}`}
+                  className="body-base-md inline-flex gap-x-2 rounded-lg bg-primary-50 px-3 py-2 text-primary-700"
+                >
+                  <Icon name="Phone" className="h-6 w-6 text-neutral-900" />
```
piotr-pajak, line 70:
```suggestion
                  <Icon name="Phone" className="size-6 text-neutral-900" />
```

apps/web/app/modules/Profile/Profile.page.tsx
```diff
@@ -0,0 +1,121 @@
...
+                  <span>{userDetails?.contactPhone}</span>
+                </a>
+                <a
+                  href={`mailto:${userDetails?.contactEmail}`}
+                  className="body-base-md inline-flex gap-x-2 rounded-lg bg-primary-50 px-3 py-2 text-primary-700"
+                >
+                  <Icon name="Email" className="h-6 w-6 text-neutral-900" />
```
piotr-pajak, line 77:
```suggestion
                  <Icon name="Email" className="size-6 text-neutral-900" />
```

piotr-pajak (commented):
GJ! I left a few suggestions for improvements.

apps/web/app/modules/Profile/Profile.page.tsx
Japrolol: We cannot use this hook here as useUserRole checks the current user while hasPermission is for the user who's profile we are viewing. Maybe I could change the name of the variable to make more sense

apps/api/src/user/user.service.ts
Japrolol: I can try it but later in the code we return the entire userBio. If userId doesn't exist, I believe instead of returning a 400, it will return 200. But I'll check to make sure

apps/api/src/user/schemas/user.schema.ts
Japrolol: I tried applying Enum, but that field appears to be a Text in database, so it says type role is not compatible. If you believe it's worth the effort, I could potentially modify the database schema. But that would probably take some time.

apps/api/src/user/user.service.ts
Japrolol: Seems removing the id causes the typing to fall apart as id is a mandatory field, I could potentially remove the typing. But not sure the performance gain is substantial enough for risking clarity and safety

apps/api/src/user/schemas/user.schema.ts
piotr-pajak: In that case, `Type.String()` seems perfectly fine to me.
@wielopolski do you think updating the schema is worth the effort?

piotr-pajak (approved):
🙌🏻 Great work! Please hold off with the merge until @wielopolski replies regarding the potential effort to change the schema.

apps/web/app/modules/Profile/Profile.page.tsx
Japrolol: I talked about that with @wielopolski, he said shared repo's aren't implemented yet in this project, and that they need to implement it ASAP, unless I misunderstood

apps/api/src/user/schemas/user.schema.ts
wielopolski: When we use casting in query to db, type in result is acceptable. You need to add sql<Type>'columnName'

wielopolski (approved):

PR #496, opened 2025-07-04
fix(SCS-483): Add missing translations and improve pie chart and charttooltips

alwoodm:
## Jira issue(s)
[SCS-483](https://github.com/Selleo/mentingo/issues/483)

## Overview
In FiveMostPopularCoursesChart fixed title of tooltip from "studentCount" to "Students" (But that need to be refactor)
Make pie charts bolder like in the figma design
Fix the design of the chart tooltip when name is long 

## Screenshots
obraz
obraz
obraz


piotr-pajak (approved):

apps/api/src/courses/course.controller.ts
```diff
@@ -251,7 +251,7 @@ export class CourseController {
   ): Promise<BaseResponse<{ id: UUIDType; message: string }>> {
     const { id } = await this.courseService.createCourse(createCourseBody, currentUserId);
 
-    return new BaseResponse({ id, message: "Course created successfully" });
+    return new BaseResponse({ id, message: "Pomyślnie utworzono kurs" });
```
DGrygierczyk, line 254: im not sure this is the correct approach. In my mind ideally we should return some key that should be translated, the BE shouldnt response with messages in Polish. Have you checked if the english version is still working with those changes?

Correct me if im wrong please @wielopolski @piotr-pajak

apps/web/app/modules/Statistics/Admin/components/FiveMostPopularCoursesChart.tsx
```diff
@@ -46,9 +46,10 @@ export const FiveMostPopularCoursesChart = ({ data, isLoading }: MostPopularCour
       fill: chartColors[index % chartColors.length],
     }));
 
-    const chartConfig = input?.reduce((config, { courseName }, index) => {
-      const key = `desktop${index === 0 ? "" : index}`;
-      config[key] = { label: courseName };
+    // TODO: Needs to be refactor
+    const chartConfig = input?.reduce((config) => {
+      const key = `studentCount`;
+      config[key] = { label: t("mostPopularCoursesView.other.students") };
```
DGrygierczyk, line 52: Im not really a fan of leaving such a todos in the code as they tend to not be fixed for a long time. Is it possible to refactor it right away? This is better approach to just leave the code in cleaner state than before we touched it

DGrygierczyk (changes requested):

PR #498, opened 2025-07-04
refactor(SCS): change h-* w-* classes to size-* when the element is square

its-gabo:
## Overview
I've changed `'h-* w-*'` classes in every place to `'size-*'` when the element was square. As we talked with @wielopolski 

Also changed every `-[1px]` class to `-px` class.


wielopolski (approved):
Good Job 🍻

PR #522, opened 2025-07-10
feat(SCS-464): implement avatar change

its-gabo:
## Issue(s)
[SCS-464](https://github.com/Selleo/mentingo/issues/464)

## Overview
This PR implements profile picture functionality for user profiles, allowing users to upload, update, and display avatar images throughout the application.

### Backend:
- **Database Schema**: Added `profilePictureS3Key` field to user schema with database migration
- **User Service**: Enhanced user service to handle profile picture uploads and updates
- **S3 Service**: Changed keys to properly handle file uploads

### Frontend
- **Profile Components**: ProfileCard and ProfileEditCard can display and manage profile pictures
- **Image Upload**: Improved ImageUploadInput component for better avatar handling
- **Profile Page**: Updated profile page to integrate profile picture functionality

### User profile with avatar
image

### User profile edit with avatar
image

### User profile edit with no avatar
image

### User with placeholder avatar
image

## Notes 
- Requires migration
- I've ensured that all endpoints returning user data exclude the `profilePictureS3Key` field


apps/api/src/user/schemas/user.schema.ts
```diff
@@ -15,6 +14,28 @@ export const userDetailsSchema = Type.Object({
   role: Type.Enum(USER_ROLES),
 });
 
+export const userWithoutProfilePictureKeySchema = Type.Omit(commonUserSchema, [
```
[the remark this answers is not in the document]
its-gabo: As we talked with @wielopolski we don't want to send `profilePictureS3Key` back to FE. We just want the url of the picture. So i needed this schema to adjust other endpoints, because they were based on `commonUserSchema`, which now includes the key that we don't want to send back.

apps/api/src/s3/s3.service.ts
```diff
@@ -15,14 +15,13 @@ export class S3Service {
 
   constructor(private configService: ConfigService) {
     this.s3Client = new S3Client({
-      endpoint: this.configService.get<string>("s3.S3_ENDPOINT") || "",
-      region: this.configService.get<string>("s3.S3_REGION") || "us-east-1",
+      region: this.configService.get<string>("AWS_REGION") || "us-east-1",
       credentials: {
-        accessKeyId: this.configService.get<string>("s3.S3_ACCESS_KEY_ID") || "",
-        secretAccessKey: this.configService.get<string>("s3.S3_SECRET_ACCESS_KEY") || "",
+        accessKeyId: this.configService.get<string>("AWS_ACCESS_KEY_ID") || "",
```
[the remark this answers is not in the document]
its-gabo: Because it didn't work with current AWS setup, and I couldn't upload a file

apps/api/src/file/file.controller.ts
```diff
@@ -23,7 +23,7 @@ import { FileUploadResponse } from "./schemas/file.schema";
 export class FileController {
   constructor(private readonly fileService: FileService) {}
 
-  @Roles(USER_ROLES.ADMIN, USER_ROLES.CONTENT_CREATOR)
+  @Roles(...Object.values(USER_ROLES))
```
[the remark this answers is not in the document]
its-gabo: I think we don't want anyone to be able to access the the file upload. Just the people with some sort of role

apps/web/app/modules/Profile/Profile.page.tsx
```diff
@@ -56,17 +57,27 @@ export default function ProfilePage() {
...
-  } = useForm<UpdateUserProfileBody>({ resolver: zodResolver(updateUserProfileSchema) });
+  } = useForm<UpdateUserProfileBody>({
+    resolver: zodResolver(updateUserProfileSchema),
+  });
 
   const { mutate: updateUserProfile } = useUpdateUserProfile();
 
   const onSubmit = (data: UpdateUserProfileBody) => {
-    if (isDirty) {
+    if (isDirty || data.profilePictureS3Key !== undefined) {
```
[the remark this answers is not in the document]
its-gabo: It's made like that, because we need to handle 3 conditions. Since we don't want to send key back to FE i have really no better way to check if user wants to change avatar, delete avatar or if nothing has changed. When we have `undefined` we know that user didn't touch it, so we don't send it to backend. If it's set to `""`, we know that user wants to delete his profile picture. And if it's a valid key, then we just know that user changed his avatar.

apps/web/app/components/FileUploadInput/ImageUploadInput.tsx
```diff
@@ -24,7 +25,7 @@ const ImageUploadInput = ({
   return (
     <div className="flex flex-col items-center justify-center gap-y-2">
       <div className="relative flex h-80 w-full cursor-pointer flex-col items-center justify-center overflow-hidden rounded-lg border-2 border-solid border-gray-300 bg-gray-100">
-        {field.value && (
+        {(!isEmpty(imageUrl) || !isEmpty(field.value)) && (
```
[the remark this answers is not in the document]
its-gabo: I kinda need it to be like that, because value of this field is my S3 key. Since I don't send key to FE, this component will not show my avatar on load, even if i have one, because i don't have the S3 key. Only when i change the avatar it will be visible.

apps/api/src/storage/migrations/0017_add_profile_picture_s3_key_to_user.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "users" ADD COLUMN "profile_picture_s3_key" varchar(200);
```
Rxbsxn, line 1: I would name it some kind of `avatar_reference` or `avatar_url`. I would avoid the vendor lock-in in names. It can be confusing to developers if we change from `s3` to `GCP storage`

apps/api/src/utils/pipes/validateMultipartPipe.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    return result;
+  }
+
+  private parseValue(value: any): any {
+    if (typeof value !== "string") {
+      return value;
+    }
+
+    if (this.isJsonString(value)) {
```
Rxbsxn, line 60:
```suggestion
    if (this.isJSONString(value)) {
```

apps/api/src/utils/pipes/validateMultipartPipe.ts
```diff
@@ -0,0 +1,110 @@
...
+  constructor(private readonly schema: T) {
+    this.validator = TypeCompiler.Compile(schema);
+  }
+
+  transform(value: any): Static<T> {
+    if (!value || typeof value !== "object") {
+      throw new BadRequestException("Invalid multipart form data");
+    }
+
+    const parsedData = this.parseMultipartToJson(value);
```
Rxbsxn, line 20:
```suggestion
    const parsedData = this.parseMultipartToJSON(value);
```

apps/web/app/modules/Profile/components/ProfileCard.tsx
Rxbsxn, line 1: I would focus on removing as many `any`as we can. Replacing with proper types based on the data we're going to send to the methods. If we are not sure, I would go with `unknown`

Rxbsxn (changes requested):
@its-gabo a few things from my side

apps/api/src/storage/migrations/0017_add_profile_picture_s3_key_to_user.sql
wielopolski: It could be good catch. `avatar_reference` could be more readable

apps/api/src/user/schemas/user.schema.ts
```diff
@@ -15,6 +14,18 @@ export const userDetailsSchema = Type.Object({
   role: Type.Enum(USER_ROLES),
 });
 
-export type UserDetails = Static<typeof userDetailsSchema>;
-export type UserResponse = Static<typeof commonUserSchema>;
+export const baseUserResponseSchema = Type.Omit(commonUserSchema, ["profilePictureS3Key"]);
+
+export const userDetailsResponseSchema = Type.Object({
+  ...userDetailsSchema.properties,
+  profilePictureUrl: Type.Union([Type.String(), Type.Null()]),
```
wielopolski, line 21: We don't return any default from BE for avatar?

apps/api/src/user/schemas/updateUser.schema.ts
```diff
@@ -19,6 +19,7 @@ export const upsertUserDetailsSchema = Type.Object({
 export const updateUserProfileSchema = Type.Object({
   firstName: Type.Optional(Type.String()),
   lastName: Type.Optional(Type.String()),
+  file: Type.Optional(Type.Union([Type.String(), Type.Null()])),
```
wielopolski, line 22: This name could be not readable for developing in the future

apps/api/src/user/user.module.ts
```diff
@@ -1,14 +1,16 @@
...
 
 import { EmailModule } from "src/common/emails/emails.module";
+import { FileService } from "src/file/file.service";
+import { S3Service } from "src/s3/s3.service";
 
 import { UserController } from "./user.controller";
 import { UserService } from "./user.service";
 
 @Module({
   imports: [EmailModule],
```
wielopolski, line 11: Where are a imports to new used services?

apps/api/src/user/user.controller.ts
```diff
@@ -164,15 +170,34 @@ export class UserController {
...
+        },
+      ],
+    },
   })
   async updateUserProfile(
-    @Body() data: UpdateUserProfileBody,
+    @Body(new ValidateMultipartPipe(updateUserProfileSchema))
+    userInfo: { data: UpdateUserProfileBody },
     @CurrentUser("userId") currentUserId: UUIDType,
+    @UploadedFile() file?: Express.Multer.File,
```
wielopolski, line 198: Call full name of file. In the future you need to be thinking if it is pdf with competencies, cv or maybe any video.
Maybe `avatarImage` or something similar

apps/api/src/user/user.service.ts
```diff
@@ -121,7 +124,17 @@ export class UserService {
     if (!canView) {
       throw new ForbiddenException("Cannot access user details");
     }
-    return userBio;
+
+    const { profilePictureS3Key, ...user } = userBio;
+
+    const profilePictureUrl = profilePictureS3Key
+      ? await this.fileService.getFileUrl(profilePictureS3Key)
```
wielopolski, line 131: Why you dont use `getSignedUrl`?

apps/api/src/user/user.service.ts
```diff
@@ -161,17 +174,29 @@ export class UserService {
...
     if (!existingUser) {
       throw new NotFoundException("User not found");
     }
 
+    if (!data && !file) {
+      throw new NotFoundException("No data provided for user profile update");
+    }
+
+    if (file && !data.file) {
+      const { fileKey } = await this.fileService.uploadFile(file, "user");
```
wielopolski, line 189:
```suggestion
      const { fileKey } = await this.fileService.uploadFile(file, "user-avatars");
```

apps/api/src/storage/schema/index.ts
```diff
@@ -22,6 +22,7 @@ export const users = pgTable("users", {
   email: text("email").notNull().unique(),
   firstName: text("first_name").notNull(),
   lastName: text("last_name").notNull(),
+  avatarReference: varchar("avatar_reference", { length: 200 }),
```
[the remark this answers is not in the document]
its-gabo: I think using `varchar` here is a better choice than `text`. I read that, it's more efficient to use varchar on the fields, that you know the length of. I've also checked in with @Rxbsxn and he said he would use `varchar` too.

apps/web/app/modules/Profile/Profile.page.tsx
```diff
@@ -56,17 +57,29 @@ export default function ProfilePage() {
...
-      updateUserProfile({ data: filteredData, id });
-      reset(data);
+      if (data.userAvatar === null) {
+        filteredData.userAvatar = null;
+      }
+
+      updateUserProfile({
+        data: { ...filteredData },
+        id,
+        userAvatar: data.userAvatar || undefined,
```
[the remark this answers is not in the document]
its-gabo: Because in `useUpdateUserProfile` the userAvatar is optional
```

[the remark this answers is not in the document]
its-gabo: Also every field is optional and that makes then `value || undefined`

arekkonopka (approved):

PR #535, opened 2025-07-14
 feat(SCS-512 | BA ): Email Notifications About New Users

Kejooorek:
## Jira issue(s)
[LC-512](https://github.com/orgs/Selleo/projects/31?pane=issue&itemId=118768937&issue=Selleo%7Cmentingo%7C512)

## Overview
I developed a backend feature to send email notifications to admins about new user registrations. Additionally, I created a settings table to store user-specific notification preferences.


apps/api/src/auth/auth.service.ts
```diff
@@ -41,8 +44,44 @@ export class AuthService {
     private emailService: EmailService,
     private createPasswordService: CreatePasswordService,
     private resetPasswordService: ResetPasswordService,
+    private settingsService: SettingsService,
   ) {}
 
+  private ensureUserSettings(user: any): CommonUser {
```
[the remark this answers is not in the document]
Kejooorek: I've just deleted this method, because It is not necessary

apps/api/src/auth/auth.service.ts
```diff
@@ -41,8 +44,44 @@ export class AuthService {
...
+    return {
+      ...user,
+      settings: (user.settings as UserSettings) || {},
+    };
+  }
+
+  private async notifyAdminsAboutNewUser(user: CommonUser) {
```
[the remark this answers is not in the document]
Kejooorek: okay, but using you approach I must define `userService`  and `emailService` inside `settingsService` because for this one function :/

BuczUu (commented):
looks good gj

apps/api/src/user/user.controller.ts
```diff
@@ -137,7 +136,8 @@ export class UserController {
         throw new ForbiddenException("You can only update your own account");
       }
 
-      const updatedUser = await this.usersService.updateUser(id, data);
+      await this.usersService.updateUser(id, data);
+      const updatedUser = await this.usersService.getUserById(id);
```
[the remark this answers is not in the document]
Kejooorek: We need to return like this because of the type mismatch  - method `getUserById` returns user with joined settings, but method `updateUser` return User without settings

DGrygierczyk (changes requested):
We need to fix the migrations too as this would introduce a lot of problems during deployment

apps/api/src/storage/migrations/0017_create_settings_table.sql
piotr-pajak, line 1: ✂️ Duplicated migration file

apps/api/src/auth/auth.service.ts
```diff
@@ -103,7 +133,7 @@ export class AuthService {
       throw new UnauthorizedException("User not found");
     }
 
-    return user;
+    return { ...user, settings: user.settings || {} };
```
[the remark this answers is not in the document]
wielopolski: Do we need to return the settings?

apps/api/src/storage/migrations/0017_create_settings_table.sql
wielopolski: This migration is correct. In code already exist migration `0016` with different name

apps/api/src/user/user.service.ts
```diff
@@ -303,6 +325,19 @@ export class UserService {
     });
   }
 
+  public async getAdminsWithSettings() {
```
wielopolski, line 328: Are we using this endpoint?

apps/api/test/factory/settings.factory.ts
```diff
@@ -0,0 +1,28 @@
...
+      const [inserted] = await db.insert(settings).values(setting).returning();
+      return inserted;
+    });
+
+    return {
+      id: faker.string.uuid(),
+      createdAt: faker.date.past().toISOString(),
+      updatedAt: faker.date.recent().toISOString(),
+      settings: DEFAULT_USER_SETTINGS,
+      userId: faker.helpers.arrayElement([faker.string.uuid(), null]),
```
wielopolski, line 25: It can be not working. We have reference to users table

apps/api/src/auth/auth.service.ts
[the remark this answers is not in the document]
Kejooorek: Yes, for type reasons, but I can think about this code below to make possible to return without settings :)

```
const userSchema = createSelectSchema(users);

export const commonUserSchema = Type.Composite([
  userSchema,
  Type.Object({
    settings: settingsSchema.properties.settings,
  }),
]);

export type CommonUser = Static<typeof commonUserSchema>;

```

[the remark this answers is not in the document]
Kejooorek: nvm - already fixed :p

apps/api/test/factory/settings.factory.ts
```diff
@@ -0,0 +1,28 @@
...
+import { DEFAULT_USER_SETTINGS } from "src/settings/constants/settings.constants";
+import { settings } from "src/storage/schema";
+
+import type { InferSelectModel } from "drizzle-orm";
+import type { DatabasePg, UUIDType } from "src/common";
+
+type SettingTest = InferSelectModel<typeof settings>;
+export type SettingsTest = SettingTest[];
+
+export const createSettingsFactory = (db: DatabasePg, userId: UUIDType) => {
```
typeWolffo, line 13: Is this factory used anywhere? From what I can see, the user factory uses the same code but already has the correct user ID.

Kejooorek: yes, it is used, I've made changes on the latest commit :)

typeWolffo: noice :shipit:

apps/api/src/settings/schemas/settings.schema.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { UUIDSchema } from "src/common";
+
+import type { Static } from "@sinclair/typebox";
+
+export const baseSettingsJSONContentSchema = Type.Object({
+  language: Type.String(),
+});
+
+export const studentSettingsJSONContentSchema = Type.Object({
```
wielopolski, line 11: It is duplicate, for the moment `studentSettingsJSONContentSchema` or `baseSettingsJSONContentSchema` will be enough

apps/api/src/settings/schemas/settings.schema.ts
```diff
@@ -0,0 +1,34 @@
...
+});
+
+export const settingsJSONContentSchema = Type.Union([
+  studentSettingsJSONContentSchema,
+  adminSettingsJSONContentSchema,
+]);
+
+export const settingsSchema = Type.Object({
+  userId: UUIDSchema,
+  settings: settingsJSONContentSchema,
```
wielopolski, line 27:
```suggestion
  settings: Type.Union([
  studentSettingsJSONContentSchema,
  adminSettingsJSONContentSchema,
]);,
```

apps/api/src/settings/schemas/settings.schema.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const adminSettingsJSONContentSchema = Type.Object({
+  ...baseSettingsJSONContentSchema.properties,
+  adminNewUserNotification: Type.Boolean(),
+});
+
+export const settingsJSONContentSchema = Type.Union([
+  studentSettingsJSONContentSchema,
+  adminSettingsJSONContentSchema,
+]);
```
wielopolski, line 23: ✂️

apps/api/src/settings/settings.service.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    if (!updated) {
+      throw new NotFoundException("User settings not found");
+    }
+
+    return updated;
+  }
+
+  public async updateAdminNewUserNotification(userId: UUIDType) {
+    const [userSettings] = await this.db.select().from(settings).where(eq(settings.userId, userId));
```
wielopolski, line 88: please select only one setting from db, by `@>` or another sql operator to JSONB

apps/api/src/storage/migrations/0019_create_settings_table.sql
```diff
@@ -0,0 +1,27 @@
...
+
+INSERT INTO settings (user_id, settings)
+SELECT u.id, 
+       CASE 
+           WHEN u.role = 'admin' THEN '{"language": "en","adminNewUserNotification": true}'::JSONB
+           ELSE '{"language": "en"}'::JSONB
+       END as settings
+FROM users u
+LEFT JOIN settings s ON u.id = s.user_id
+WHERE s.user_id IS NULL;
```
wielopolski, line 27: You can add default setting for whole app (without userId)

apps/api/src/user/user.service.ts
```diff
@@ -304,6 +314,26 @@ export class UserService {
     });
   }
 
+  public async getAdminsToNotifyAboutNewUser() {
+    const allAdmins = await this.db
```
wielopolski, line 318: I'm not sure, but notifications are supposed to be a global setting, not for a specific user with the admin role. To check with @maciejhubert

wielopolski (commented):
Looks much better 🍻 , I left some suggestions

apps/api/src/user/user.service.ts
Kejooorek: You can be right, maybe I misunderstood something:
image

apps/api/src/settings/settings.service.ts
```diff
@@ -0,0 +1,139 @@
...
+      .returning();
+
+    return updated;
+  }
+
+  private getDefaultSettingsForRole(role: UserRole): SettingsJSONContentSchema {
+    switch (role) {
+      case USER_ROLES.ADMIN:
+        return {
+          language: "en",
```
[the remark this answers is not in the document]
Kejooorek: done :)

PR #543, opened 2025-07-17
fix(SCS-540): made lesson container scrollable

Japrolol:
## Jira issue(s)
[SCS-540](https://github.com/Selleo/mentingo/issues/540)

## Overview
Modified min-h-dvh to max-h-dvh so all overflow is handled inside the full viewport height. Also made body overflow hidden due to the fact that there were issues were you would scroll outside of body.

## Screenshots
https://github.com/user-attachments/assets/c67f2a1c-2a19-419a-851a-c53a94743eda

### DIFF
--- a/apps/web/app/modules/Courses/Lesson/LessonSidebar.tsx
+++ b/apps/web/app/modules/Courses/Lesson/LessonSidebar.tsx
@@ -45,7 +45,7 @@ export const LessonSidebar = ({ course, lessonId }: LessonSidebarProps) => {
   if (!course) return null;
 
   return (
-    <div className="h-full w-full rounded-lg bg-white">
+    <div className="sticky top-0 h-auto max-h-fit min-h-screen overflow-y-scroll rounded-lg bg-white">
       <div className="flex flex-col gap-y-12">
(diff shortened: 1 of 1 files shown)


piotr-pajak (approved):

piotr-pajak:
Please check if the change in the body doesn't affect the rest of the application. 😉

apps/web/app/root.tsx
```diff
@@ -30,7 +30,7 @@ export function Layout({ children }: { children: React.ReactNode }) {
         <Meta />
         <Links />
       </head>
-      <body>
+      <body className="overflow-hidden">
```
Rxbsxn, line 33: I would move this from <body> tag, to some container element. I wouldn't keep such an “impacting” thing in one of the most important HTML blocks in the application

arekkonopka: I also wouldn't add this styling to the `body` tag, if it work's today it could break something in the future. Please reconsider this approach, it should be different.

Japrolol: Sorry, I forgot to push changes. I removed overflow hidden from body tag.

PR #562, opened 2025-07-24
fix(SCS-551): user avatar display

its-gabo:
## Jira issue(s)
[LC-]()

### Notes 

- [ ] fired up e2e tests and got a positive result


apps/web/app/components/UserProfile/UserAvatar.tsx
```diff
@@ -0,0 +1,22 @@
...
+      <AvatarImage
+        src={profilePictureUrl}
+        alt={`${userName} profile`}
+        className="h-full w-full object-cover"
+      />
+    ) : (
+      <Gravatar email="email@example.com" />
+    )}
+  </Avatar>
+);
```
piotr-pajak, line 22: WDYT 🤔 
```suggestion
export const UserAvatar = ({ email, userName, profilePictureUrl, className }: UserAvatarProps) => (
  <Avatar className={className}>
    {profilePictureUrl ? (
      <AvatarImage
        src={profilePictureUrl}
        alt={`${userName} profile`}
        className="h-full w-full object-cover"
      />
    ) : (
      <Gravatar email={email} />
    )}
  </Avatar>
);

```

its-gabo: This could get out of hand quickly. I’d stick with the current approach, since we can't be sure what Gravatar will set as the profile picture. With our current method, we guarantee a consistent profile image for every user that doesn’t change unexpectedly.

PR #566, opened 2025-07-24
feat(SCS-526): [BE] Ability to customize platform logo

alwoodm:
## Jira issue(s)
[SCS-526](https://github.com/Selleo/mentingo/issues/526)

## Overview
Added platform logo management system. Administrators can upload custom logos (PNG/SVG, max 2MB). Logo is stored in AWS S3 and accessible through public API endpoint for display on login page and sidebar. Falls back to default Mentingo logo when no custom logo is set.


apps/api/src/settings/settings.service.ts
```diff
@@ -121,6 +130,69 @@ export class SettingsService {
     return updated;
   }
 
+  public async getGlobalSettings() {
+    const [globalSettings] = await this.db
```
Kejooorek, line 134: maybe something like this?

```
    const [globalSettings] = await this.db.select().from(settings).where(isNull(settings.userId));
```

apps/api/src/settings/settings.controller.ts
```diff
@@ -39,4 +52,25 @@ export class SettingsController {
     const result = await this.settingsService.updateAdminNewUserNotification(userId);
     return new BaseResponse(result);
   }
+
+  @Get("platform-logo")
+  @Public()
+  @Validate({
+    response: baseResponse(Type.Object({ url: Type.Union([Type.String(), Type.Null()]) })),
```
Kejooorek, line 59: maybe wrap it (`Type.Object({ url: Type.Union([Type.String(), Type.Null()])`)  as a  variable and then use it here?

apps/api/src/settings/settings.service.ts
```diff
@@ -121,6 +130,69 @@ export class SettingsService {
...
+
+      return created;
+    }
+
+    return globalSettings;
+  }
+
+  public async updateGlobalSettings(updatedSettings: Partial<GlobalSettingsType>) {
+    const currentSettings = await this.getGlobalSettings();
+    const currentSettingsData = currentSettings.settings as any;
```
Kejooorek, line 157: why `any` here?

apps/api/src/settings/constants/settings.constants.ts
```diff
@@ -6,3 +6,11 @@ export const DEFAULT_USER_ADMIN_SETTINGS = {
   ...DEFAULT_USER_SETTINGS,
   adminNewUserNotification: false,
 };
+
+export const DEFAULT_GLOBAL_SETTINGS = {
+  platformLogoS3Key: null as string | null,
+} as const;
+
+export type GlobalSettingsType = {
```
arekkonopka, line 14: Types should be kept in types file.

apps/api/src/settings/settings.controller.ts
arekkonopka, line 59: Please create schema.

apps/api/src/settings/settings.service.ts
```diff
@@ -121,6 +130,69 @@ export class SettingsService {
     return updated;
   }
 
+  public async getGlobalSettings() {
+    const [globalSettings] = await this.db
+      .select()
+      .from(settings)
+      .where(sql`user_id IS NULL`);
```
arekkonopka, line 137: Please use `isNull` https://orm.drizzle.team/docs/operators#isnull

apps/api/src/settings/settings.service.ts
```diff
@@ -121,6 +130,69 @@ export class SettingsService {
...
+
+    if (!globalSettings) {
```
arekkonopka, line 139: Why do we create globalSettings here? Shouldn't it be solved in other way? Maybe with seeds or something like this? 
cc: @wielopolski

apps/api/src/settings/settings.service.ts
arekkonopka: Please use correct type instead of `any` or `unknown` if you don't know the type.

apps/api/src/settings/constants/settings.constants.ts
arekkonopka: It's not needed to have `type` word in it, and also i think it should consist `JSON` in name, since it is JSON

apps/api/src/settings/settings.controller.ts
```diff
@@ -39,4 +52,25 @@ export class SettingsController {
...
+  async getPlatformLogo() {
+    const url = await this.settingsService.getPlatformLogoUrl();
+    return new BaseResponse({ url });
+  }
+
+  @Patch("platform-logo")
+  @Roles(USER_ROLES.ADMIN)
+  @UseInterceptors(FileInterceptor("logo"))
+  async updatePlatformLogo(@UploadedFile() logo: Express.Multer.File): Promise<void> {
+    if (!logo) {
```
arekkonopka, line 70: please move it to service

apps/api/src/file/file.service.ts
```diff
@@ -20,34 +20,53 @@ export class FileService {
...
       throw new BadRequestException(
         `File size exceeds the maximum allowed size of ${MAX_FILE_SIZE} bytes`,
       );
     }
 
-    if (!allowedMimeTypes.includes(file.mimetype)) {
+    let mimetype: string | undefined = file.mimetype;
+    if (!mimetype) {
+      const extension = file.originalname.split(".").pop()?.toLowerCase();
+      const mimeTypeMap: Record<string, string> = {
```
Kejooorek, line 46: maybe move this code to a separate file, maybe something like `.consts`

apps/api/src/settings/settings.service.ts
Kejooorek: I've implemented functionality for creating settings using seeds (for development purposes) and migrations that are responsible for creating settings for each user (based on role) plus global settings

wielopolski (commented):
Good job, I left some suggestion

apps/api/src/settings/settings.service.ts
```diff
@@ -121,6 +133,72 @@ export class SettingsService {
     return updated;
   }
 
+  public async getGlobalSettings() {
+    const [globalSettings] = await this.db.select().from(settings).where(isNull(settings.userId));
```
wielopolski, line 137: It would be good to pull out the settings right away we don't need the rest and type it in the query right away

arekkonopka (approved):

apps/api/src/file/guards/mime-type.guard.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { ALLOWED_MIME_TYPES } from "../file.constants";
+
+export class MimeTypeGuard {
+  static validateMimeType(mimetype: string | undefined): asserts mimetype is string {
+    if (!mimetype || !this.isAllowedMimeType(mimetype)) {
+      throw new BadRequestException(
+        `File type ${
+          mimetype || "unknown"
+        } is not allowed. Allowed types are: ${ALLOWED_MIME_TYPES.join(", ")}`,
```
Rxbsxn, line 11: Is the error translation handled in frontend?

alwoodm: Frontend form validation prevents users from selecting invalid file types, so this backend error only occurs for direct API calls and should remain as technical error message in English.

PR #620, opened 2025-09-01
Ak feat course 609 add new status to courses

arekkonopka:
## Jira issue(s)
[LC-]()

### Notes 

- [ ] fired up e2e tests and got a positive result


wielopolski (approved):
LGTM 🍻 
I would consider adding constant with these values, maybe even StatusEnum or Status  for more general use.

```
const COURSE_STATUS = {
  draft: "draft",
  published: "published",
  private: "private",
} as const;

type CourseStatus = typeof COURSE_STATUS[keyof typeof COURSE_STATUS];

PR #647, opened 2025-09-16
fix: Incorrect behavior when combining SSO with MFA login

its-gabo:
## Overview
- Fixed setting up MFA after SSO login if required.
- Deleted unnecessary `.env` variable
- Fixed tests that failed due to change


Jlkan (approved):
lgtm 👍

PR #659, opened 2025-09-22
feat: allow admin to change currency of course price

its-gabo:
## Issue(s)
[656](https://github.com/Selleo/mentingo/issues/656)

## Overview
- Implemented `Select` for changing default course currency by admin. 

## Screenshots
https://github.com/user-attachments/assets/37ec2a08-d3ca-4aad-9fd2-d2bc24b1f9de

https://github.com/user-attachments/assets/55684ade-a0aa-46d9-90ab-d06b223938f9


maciejhubert:
@its-gabo I know we spoke that for simplicity we'll have the symbol before the value, but when I saw it in the button here it looks really weird:
image

Let's add 3 more improvements:
* PLN is displayed as `zł`
* `zł` symbol is displayed after the value
* let's truncate the value if there are only zeros after dot

So in the case above `123 zł` should be displayed.

its-gabo:
Enhanced currency formatting for different locales.

https://github.com/user-attachments/assets/3f007462-8a9a-4f51-8dc9-83e3c06ccac1

PR #686, opened 2025-09-25
feat: implement slack authentication

its-gabo:
## Jira issue(s)
[667](https://github.com/Selleo/mentingo/issues/667)

## Overview
Implemented slack authentication

## Screenshots

https://github.com/user-attachments/assets/0d96f5e7-0114-4fde-8263-ce2f27f2a40e

### Notes 
To test this locally you need to have a bit changed setup. You'll need to have a tunnel to your frontend and change envs.
For more info @its-gabo 

! Slack app has to have configured scopes correctly
image


apps/api/src/auth/strategy/slack.strategy.ts
```diff
@@ -0,0 +1,44 @@
...
+      headers: { Authorization: `Bearer ${accessToken}` },
+    });
+
+    const user = response.data.user;
+
+    return {
+      firstName: user.name,
+      lastName: "",
+      email: user.email,
+    };
```
Rxbsxn, line 42: Just try names with middle name

its-gabo: We talked about this with @maciejhubert and in the future we'll probably be implementing some kind of modal if not all of the data is provided from providers login.

PR #707, opened 2025-10-01
chore: adjust gap fills and fill in the blanks creation to be more intuitive

its-gabo:
## Issue(s)
[613](https://github.com/Selleo/mentingo/issues/613)

## Overview
Adjusted `Gap fills` and `Fill in the blanks` to figma design

## Screenshots
image

### DIFF
--- a/apps/web/app/locales/en/translation.json
+++ b/apps/web/app/locales/en/translation.json
@@ -1012,7 +1012,7 @@
           "presentationLessonDescription": "Choose this type if you want to embed presentations files like PPTX.",
           "quizLessonDescription": "This type allows you to build quizzes for your chapter using different types of questions.",
           "aiMentorLessonDescription": "Create interactive AI-powered lessons where students can engage with a virtual mentor.",
-          "fillInTheBlanksDescription": "This question will have sentence and students will need to drag and drop the word in to right place.",
+          "fillInTheBlanksDescription": "This question contains a sentence; students must drag and drop the correct word into the blank.",
           "gapFillDescription": "This question will have sentence with missing words that students need to type in.",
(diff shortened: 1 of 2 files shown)


apps/web/app/modules/Admin/EditCourse/CourseLessons/NewLesson/QuizLessonForm/components/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const FILL_IN_THE_BLANKS_BUTTON_CLASSNAME =
```
wielopolski, line 1: I'm not sure, but don't we want to keep things like that in `index.css`?

wielopolski (commented):
LGTM, I left suggestion

apps/web/app/modules/Admin/EditCourse/CourseLessons/NewLesson/QuizLessonForm/components/constants.ts
its-gabo: As we talked i think people will rather open the file with `const` than `index.css`, therefore possibility of people just doing `class="fill-in-the-blanks px-2"` will be smaller and then it will result in less situations with not working fill in the blanks will be smaller. (form validation runs on classes in this case)

maciejhubert:
@its-gabo sorry if I was vague about the requirements 😅 . We also need to:
* change the icons according to [this frame](https://www.figma.com/design/Hdnr93xRJ94FYgLr8jDICM/LMS--Mentingo-?node-id=5901-12521&t=VoQkh6YJlfnP38n0-1)
* add short explanations:
image

Issue #771, opened 2025-10-13
Fix statistics on dashboard


maciejhubert:
Currently statistics are not displayed correctly. This is an example on demo.mentingo.com which has a few users, courses and submitted answers in quizzes.

<img width="1670" height="888" alt="Image" src="https://github.com/user-attachments/assets/53501916-33fa-46cb-81d7-65dd3a107de6" />

Please fix the data that is displayed there (for both admin and student).

PR #796, opened 2025-10-15
fix: allow admin to see all statistics regardless of the course creator

zielvna:
## Jira issue(s)
[#771](https://github.com/Selleo/mentingo/issues/771)

## Overview

- Add admin only entrypoint (`GET /api/statistics/admin-stats`) to see statistics for all courses instead of only ones you created
- Update translations

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -17,6 +17,7 @@
 **/coverage
 **/*.generated.*
 **/generated
+**/generated-api.ts
 **/api/src/swagger/*
 **/.docusaurus
 **/*.d.ts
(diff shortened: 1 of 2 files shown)


apps/web/app/api/queries/useStatistics.ts
```diff
@@ -0,0 +1,33 @@
...
+    queryFn: async () => {
+      const response = await ApiClient.api.statisticsControllerGetAdminStats();
+
+      return response.data;
+    },
+    select: (data: GetAdminStatsResponse) => data.data,
+  };
+};
+
+export function useStatistics(type: string) {
```
wielopolski: What do you think about using a single endpoint and selecting the appropriate method in the service based on the current user role in the controller?
And when using constant values, use the ones already defined or define new ones as needed. For example, USER_ROLE from apps/web/app/config/userRoles.ts, or even move it to shared packages and use it in both apps.

PR #814, opened 2025-10-21
feat: set default interface language according to browser settings

RyczekMateusz:


maciejhubert:
@RyczekMateusz I cannot make it work, below you can find my attempt. I believe in this case the default UI language in Mentingo should be set to Polish.

https://github.com/user-attachments/assets/f0b3f939-5a69-4d52-94f4-76439fdbf340

PR #818, opened 2025-10-22
fix: allow author to preview private/draft courses

RyczekMateusz:


apps/api/src/courses/course.service.ts
```diff
@@ -497,7 +497,7 @@ export class CourseService {
     const NON_PUBLIC_STATUSES = ["draft", "private"];
```
igorjuraszek:
```js
const NON_PUBLIC_STATUSES = {
  draft: "draft",
  private: "private",
};
```

apps/api/src/courses/course.service.ts
```diff
@@ -497,7 +497,7 @@ export class CourseService {
...
 
     if (!course) throw new NotFoundException("Course not found");
-    if (NON_PUBLIC_STATUSES.includes(course.status) && !isEnrolled)
+    if (userId !== course.authorId && NON_PUBLIC_STATUSES.includes(course.status) && !isEnrolled)
```
igorjuraszek:
```js
if (
  userId !== course.authorId &&
  Object.values(NON_PUBLIC_STATUSES).includes(course.status) &&
  !isEnrolled
) {
  throw new ForbiddenException("You have no access to this course");
}
```

RyczekMateusz: Thanks for your comments @igorjuraszek but I don't see why we should do it this way. Can you elaborate why?

PR #828, opened 2025-10-24
refactor: improve student deletion system

Japrolol:
## Jira issue(s)
[824](https://github.com/Selleo/mentingo/issues/824)

## Overview
- Created a deletedAt timestamp in db for soft deletion
- Added delete user with basic deflation of statistics and anonimization
- Added deletedAt checks in queries where needed

## Screenshots
https://github.com/user-attachments/assets/997c8512-222c-4fb4-b24c-022407be0cd0

## Notes
- Please review this PR carefully, in case I missed a place.
- I only put deletedAt checks where actually needed, I didn't place it in each select, if the method where it was used was guarded against user not found.
- Do not merge yet, 2 statistics aren't deflated because there is no easy way to reconstruct the data, how we act further is a business decision


wielopolski (commented):
The code looks good, but I wonder about the benefit of checking everywhere whether the user has been deleted. Shouldn't this be done in key places instead of in every query? I wonder how to maintain this in further development?

apps/api/src/user/user.service.ts
```diff
@@ -346,7 +365,10 @@ export class UserService {
   }
 
   async resetPassword(id: UUIDType, newPassword: string) {
-    const [existingUser] = await this.db.select().from(users).where(eq(users.id, id));
+    const [existingUser] = await this.db
+      .select()
+      .from(users)
+      .where(and(eq(users.id, id), isNull(users.deletedAt)));
```
wielopolski: Extract this query into a separate method and use it elsewhere, please. We have few method with this query, wdyt?

wielopolski:
@maciejhubert Good point. However, I think that when such a course of action is permitted for the administrator, it is difficult to talk about an error or a loophole in the system. It will then be a conscious process, and I don't know if it's worth securing it.

Japrolol:
> The code looks good, but I wonder about the benefit of checking everywhere whether the user has been deleted. Shouldn't this be done in key places instead of in every query? I wonder how to maintain this in further development?

I blocked everywhere to not allow admin to access the deleted user in any way. But I guess I could in theory remove a few of the checks, what do you think?

maciejhubert:
> @maciejhubert Good point. However, I think that when such a course of action is permitted for the administrator, it is difficult to talk about an error or a loophole in the system. It will then be a conscious process, and I don't know if it's worth securing it.

I even checked that scenario, the only issue I found is that all courses created by such a user have a `deleted user` information as the owner. The app does not crash though, which is good enough for me to proceed with these changes :).

So after the code review feedback is applied I believe we can merge it.

maciejhubert:
@Japrolol please let me know what is the status, I see some feedback from Michal and there is a note in description:
> Do not merge yet, 2 statistics aren't deflated because there is no easy way to reconstruct the data, how we act further is a business decision

Japrolol:
> @Japrolol please let me know what is the status, I see some feedback from Michal and there is a note in description:
> 
> > Do not merge yet, 2 statistics aren't deflated because there is no easy way to reconstruct the data, how we act further is a business decision
> 
> 
> 

You can merge it

PR #863, opened 2025-10-31
fix(831): redirecting to profile page after login

RyczekMateusz:


maciejhubert:
Works well, let's fix conflicts and will merge it

PR #892, opened 2025-11-05
fix: incorrect display of course description in continue learning section

its-gabo:
## Issue
[891](https://github.com/Selleo/mentingo/issues/891)

## Overview
Fixed display of course description in continue learning card.

### Before
image

### After
image

### Notes 

Screenshot kinda bolds the text, so in reality it's kinda narrower.

- [x] Fired up e2e tests and got a positive result


maciejhubert:
works OK, let's fix conflicts @its-gabo

PR #894, opened 2025-11-05
fix: payment modal not opening on enroll click

its-gabo:
## Overview
Fixed payment modal not opening, because of not provided stripe's publishable key 

### Notes 

- [x] Fired up e2e tests and got a positive result


apps/api/src/env/services/env.service.ts
```diff
@@ -104,4 +104,12 @@ export class EnvService {
       slack,
     };
   }
+
+  async getStripePublishableKey() {
+    const stripePublishableKey = await this.getEnv("VITE_STRIPE_PUBLISHABLE_KEY")
+      .then((r) => r.value)
```
Rxbsxn:
```suggestion
      .then(({ value }) => value)
```

PR #896, opened 2025-11-05
fix: add option to drop files

RyczekMateusz:
https://github.com/Selleo/mentingo/issues/883

Implement react-dropzone for drag nad drop file functionality


maciejhubert:
The flow does not work - I can see a preview, but apparently the file was not uploaded.

After drag&drop: 

image

After attempting to save:

image

its-gabo:
Just one minor issue, when we override the video in the lesson, the old one is still displayed probably due to lack of invalidation.

PR #920, opened 2025-11-12
feat: unregistered users ability to go through lessons

DawidHaratyk:
## Issue
https://github.com/Selleo/mentingo/issues/916

## Overview
As in task
- waiting for a business decision about how we do want to tackle that, also we do have other bigger priorities that we need to tackle first, we'll come back to this task

### DIFF
--- a/apps/web/app/Guards/RouteGuard.tsx
+++ b/apps/web/app/Guards/RouteGuard.tsx
@@ -3,20 +3,33 @@ import { useLayoutEffect } from "react";
 
 import { routeAccessConfig } from "~/config/routeAccessConfig";
 import { useUserRole } from "~/hooks/useUserRole";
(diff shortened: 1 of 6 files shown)


apps/web/app/Guards/RouteGuard.tsx
```diff
@@ -32,8 +45,16 @@ export const checkRouteAccess = (path: string, userRole: UserRole) => {
       return segment === pathSegments[index];
     });
 
-    if (matches && roles.includes(userRole)) {
-      return true;
+    if (isLoggedIn) {
+      if (matches && roles.includes(userRole)) {
```
Rxbsxn: Is the check for `roles.includes` necessary here? I believe this case is handled above? What do you think?

Rxbsxn (changes requested):
@DawidHaratyk a few things from my side

PR #926, opened 2025-11-13
feat(909): allow to add images to text lessons

Japrolol:
## Issue(s)
[909](https://github.com/Selleo/mentingo/issues/909)

## Overview
- Added possibility to drag and drop, paste and upload files through toolbar in text lessons
- Added server-side validation for filetype
- User sees `<a>` in edit view and `` in lesson view. 
- Added an endpoint to stream files from S3
- Added an endpoint to add images to lesson as resources.

## Screenshots
https://github.com/user-attachments/assets/a1fe78c9-c9eb-4367-ba2d-3657c4436118

### DIFF
--- a/apps/api/src/lesson/lesson.controller.ts
+++ b/apps/api/src/lesson/lesson.controller.ts
@@ -318,7 +318,7 @@ export class LessonController {
     @Body("lessonId") lessonId: UUIDType,
     @UploadedFile() file: Express.Multer.File,
   ) {
-    return await this.adminLessonsService.uploadFileToLesson(lessonId, file);
+    return this.adminLessonsService.uploadFileToLesson(lessonId, file);
   }
 
   @Delete("delete-student-quiz-answers")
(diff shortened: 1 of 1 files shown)


apps/api/src/file/file.service.ts
```diff
@@ -92,6 +92,14 @@ export class FileService {
     }
   }
 
+  async getFileStream(fileKey: string) {
+    try {
+      return await this.s3Service.getFileStream(fileKey);
+    } catch (error) {
+      throw new ConflictException("Failed to retrieve file");
```
Rxbsxn: I would return `BadRequest` error  here

apps/api/src/lesson/lesson.controller.ts
```diff
@@ -285,6 +292,35 @@ export class LessonController {
...
+  @ApiResponse({
+    status: 201,
+    description: "File uploaded successfully",
+    type: String,
+  })
+  async uploadImageToLesson(
+    @Body("lessonId") lessonId: UUIDType,
+    @UploadedFile() file: Express.Multer.File,
+  ) {
+    return await this.adminLessonsService.uploadFileToLesson(lessonId, file);
```
Rxbsxn:
```suggestion
    return this.adminLessonsService.uploadFileToLesson(lessonId, file);
```

Issue #888, opened 2025-11-04
Implement new designs for email notifications


maciejhubert:
Here you can find [designs for email notifications](https://www.figma.com/design/raFaZ3oYEBzO3aLYYHge5U/Mentingo?node-id=2894-4516&t=qpjoyco1PTttxPxW-4). Please apply them.

Notes:
* apply designs with `blue background` and `All Rounded Gothic` (if it is not displayed properly for some reason, let's apply the `Arial Rounded MT Bold`) 
* not all courses have certificates -> for such courses:
   * `PL`: change `Ukończyłeś(-aś) {courseName}. Certyfikat jest gotowy do pobrania; sprawdź też proponowane ścieżki dalszej nauki.` to `Ukończyłeś(-aś) {courseName}. Możesz wrócić do niego w każdej chwili.`
   * `PL`: change the button to `Wróć do kursu` (and it's behavior accordingly)
   * `EN`: change `You’ve completed {courseName}. Your certificate is ready to download; check the recommended next steps.` to `You’ve completed {courseName}. You can get back to the course anytime.`
   * `EN`:  change the button to `Get back to the course` (and it's behavior accordingly)
* language of the email should be taken from settings of the user who receives the email
* the logo Microsoft logo displayed on the designs should be replaced by the uploaded company logo (if the logo is not uploaded, then display Mentingo logo)
* the title of each email should correspond to the heading:
<img width="1441" height="892" alt="Image" src="https://github.com/user-attachments/assets/326266bb-484f-4bfc-877b-1fa8016b8ab5" />

PR #822, opened 2025-10-23
feat(764): enhance range of email notification triggers

Japrolol:
## Jira issue(s)
[764](https://github.com/Selleo/mentingo/issues/764)

## Overview
- Created a user notification event handler
- Created events for notifications
- Created email templates
- Created settings for those triggers
- Created frontend for those triggers

## Screenshots
https://github.com/user-attachments/assets/04050de3-e6c6-45ae-b660-783ffa00a4ea

PR #930, opened 2025-11-14
feat: enhance email notifications for clarity and visual appeal

its-gabo:
## Issue(s)
[#888](https://github.com/Selleo/mentingo/issues/888)

## Overview
Implemented new designs for email notifications. 

Created `BaseEmailTemplate`, so the styles are the same for all emails.

### Not customized
image

### Fully customized
image

## Notes

Requires [#822] to be merged.

### DIFF
--- a/apps/api/src/user/handlers/notify-users.handler.ts
+++ b/apps/api/src/user/handlers/notify-users.handler.ts
@@ -101,14 +101,15 @@ export class NotifyUsersHandler implements IEventHandler {
 
   async notifyUserAboutInvite(event: UserInviteEvent) {
     const { userInvite } = event;
-    const { email, createdByUserName, token } = userInvite;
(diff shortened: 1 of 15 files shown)


apps/api/src/user/handlers/notify-users.handler.ts
```diff
@@ -172,17 +182,19 @@ export class NotifyUsersHandler implements IEventHandler {
     const recentCourses = await this.getRecentCourses(users);
 
     await Promise.all(
-      users.map((user) => {
+      users.map(async (user) => {
```
Rxbsxn: Do we care about order of the promises?

its-gabo: Not really, we just send it out and the order doesn't really matter as long as everyone gets their email

PR #940, opened 2025-11-14
fix: added error message for lesson title length limit

Japrolol:
## Issue(s)
[936](https://github.com/Selleo/mentingo/issues/936)

## Overview
- Added a constant for max lesson length
- Added checks with error messages for max length


its-gabo:
Maybe we could translate it? 🤔

PR #1089, opened 2025-12-12
feat: enable automatic translations powered by AI

Japrolol:
## Issue(s)
[943](https://github.com/Selleo/mentingo/issues/943)

## Overview 

- Added translation for missing fields in a course

## Business Value

- Improves content creator experience, because instead of translating everything and then checking if all is correct. Now the content creator only has to verify that everything is translated, which saves precious time.

## Screenshots / Video

https://github.com/user-attachments/assets/962263fb-8990-46cf-b213-280486519fd1

### Notes 
- It's not perfect yet, it has issues when translating language courses (some stuff we wouldn't want to translate, because that would be pointless in language courses), but acceptable for v1

### DIFF
--- a/apps/api/src/env/env.controller.ts
+++ b/apps/api/src/env/env.controller.ts
@@ -16,6 +16,7 @@ import {
   getEnvResponseSchema,
   stripePublishableKeyResponseSchema,
   isEnvSetupResponseSchema,
+  aiConfiguredResponseSchema,
 } from "src/env/env.schema";
 import { EnvService } from "src/env/services/env.service";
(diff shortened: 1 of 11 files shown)


marektro:
I was able to get this error again, here are more details:
Screenshot 2025-12-16 at 14 47 27
image

Japrolol:
> @Japrolol This happened yesterday with the initial seed data and I tried to do it 5-10 times and reloading the app, every time with the same error. Today when I am testing the same case with the same version of the app with no changes, now it works. I haven't changed anything, yesterday it was throwing errors, today it works.

Decreased chunk size, so it gets less information at once, should work better now

PR #1090, opened 2025-12-15
feat: improved video upload experience

Japrolol:
## Issue(s)
[1072](https://github.com/Selleo/mentingo/issues/1072)

## Overview 

- Allows you to navigate through app, even when files are uploading
- Implemented socket connection with auth for secure real-time bidirectional communication, between the backend and frontend, which ensures only the expected recipient gets the data.
- Implemented toasts for state messages based on data from socket
- Increased max upload size to 5gb
- Added simple reference counting to global socket to ensure it is not deleted, if something else is using it.

## Business Value

- It allows user to use app, when the video is still processing
- User gets a notification when the video is ready, so he doesn't need to check all the time.
- It allows users to upload larger videos, so they don't have to split them to chunks.

## Screenshots / Video

https://github.com/user-attachments/assets/abacdffa-5ba2-4dfc-aa98-53935b2968d7


apps/api/src/file/file.service.ts
```diff
@@ -347,4 +408,75 @@ export class FileService {
...
+  async handleBunnyWebhook(payload: BunnyWebhookBody & Record<string, unknown>) {
+    const status = payload.status ?? payload.Status ?? 0;
+
+    const videoId =
+      payload.videoId ||
+      payload.VideoId ||
+      payload.videoGuid ||
+      payload.VideoGuid ||
+      payload.guid ||
+      (payload as Record<string, string>)?.["Guid"];
```
its-gabo: Isn't that consistent? 

I've read in the documentation that it should be `VideoGuid`

its-gabo (approved):
Looks good for me 🍻

k1eu:
How does it work in context of the new Combined lesson - when video is part of the bigger lesson - not a separate one?
Shouldn't we indicate - that something is happening in the background? Some fake progress bar with label that it is uploading? 
On the Video Lesson instead of endlessly loading - shouldn't we show that the video is uploading / processing?

PR #1140, opened 2025-12-23
feat: introduce active user count analytics

Japrolol:
## Issue(s)
[1104](https://github.com/Selleo/mentingo/issues/1104)

## Overview 

- Implemented new controller with analytics routes for usage tracking. Endpoints are guarded with a .env API_KEY, which is extracted from request headers.
- The endpoint tracks active users, by checking, whether they have ever logged in and if their account is not deleted.

## Business Value

# Video showing the complete working flow of the feature / enhancement / bug-fix included in this code change.
-->

https://github.com/user-attachments/assets/5bb1585a-9636-4e3b-b682-cb425985906d

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -81,3 +81,7 @@ MASTER_KEY=
 # TEST
 DATABASE_TEST_URL=postgresql://postgres:guidebook@localhost:5432/guidebook_test
 REDIS_TEST_URL=redis://localhost:6379/1
+
+
+# Analytics secret (generate: openssl rand --base64 32)
+ANALYTICS_SECRET=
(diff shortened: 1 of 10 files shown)


apps/api/.env.example
```diff
@@ -81,3 +81,7 @@ MASTER_KEY=
 # TEST
 DATABASE_TEST_URL=postgresql://postgres:guidebook@localhost:5432/guidebook_test
 REDIS_TEST_URL=redis://localhost:6379/1
+
+
+# Analytics secret (generate: openssl rand --base64 32)
+ANALYTICS_SECRET=
```
PoslinskiNet: Is it required?

PoslinskiNet:
Let's merge it and integrate it with our excel so it automatically fetches numbers of users on each instance automatically @its-gabo [generate via gpt api call as a custom function in google sheets]

its-gabo (approved):

PR #1145, opened 2025-12-23
feat: allow admins to define a minimum age for registration

staskowsky:
## Issue(s)
[1125](https://github.com/Selleo/mentingo/issues/1125)

## Overview  
Implemented age limitation feature. In settings, admin can add a select field to the register form. This field validates, if user is of given age or older. 

## Screenshots / Video
image
image

### DIFF
--- a/apps/api/src/settings/schemas/settings.schema.ts
+++ b/apps/api/src/settings/schemas/settings.schema.ts
@@ -45,9 +45,7 @@ export const globalSettingsJSONSchema = Type.Object({
   newsEnabled: Type.Boolean(),
   unregisteredUserArticlesAccessibility: Type.Boolean(),
   articlesEnabled: Type.Boolean(),
-  ageLimit: Type.Union(
-    ALLOWED_AGE_LIMITS.map((currency) => (!currency ? Type.Null() : Type.Literal(currency))),
-  ),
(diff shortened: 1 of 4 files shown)


apps/api/src/settings/constants/settings.constants.ts
```diff
@@ -53,3 +53,4 @@ export const DEFAULT_ADMIN_SETTINGS = {
 };
 
 export const ALLOWED_CURRENCIES = ["pln", "eur", "gbp", "usd"] as const;
+export const ALLOWED_AGE_LIMITS = [13, 16, null] as const;
```
its-gabo: I've seen that you have it declared on both frontend and backend.

You could move it to `packages/shared` to have it declared in one place. Therefore we will not have to worry about 2 places, when we need to add an option 😉

apps/web/app/modules/Auth/Register.page.tsx
```diff
@@ -78,8 +83,39 @@ export default function RegisterPage() {
...
+        setAgeValidationError(
+          t(`registerView.validation.birthday`, { ageLimit: globalSettings.ageLimit }),
+        );
+      } else {
+        setAgeValidationError(null);
+      }
+    } else {
+      setAgeValidationError(null);
+    }
+  }, [birthdayValue, globalSettings?.ageLimit, t]);
```
its-gabo: I'm pretty sure that we could do the error handling using using zod's `superRefine`

PR #1170, opened 2026-01-07
fix: ensure release process runs correctly

its-gabo:
## Overview
Remove an unnecessary trailing backslash in the Makefile release target to avoid a redundant line continuation.

## Business Value
Keeps the release script clean and reduces the chance of subtle shell parsing issues during release automation.


maciejhubert (approved):
LGTM

Issue #1149, opened 2025-12-29
Improve experience for mandatory courses


maciejhubert:
Let's improve experience for mandatory courses.

### Issue 1:

Missing tooltip:

<img width="1040" height="1016" alt="Image" src="https://github.com/user-attachments/assets/363de4df-8708-4dfd-a281-eabdd12dd3ec" />

Text
* PL: `Oznacz kurs jako obowiązkowy dla tej grupy. Wymagane jest wskazanie daty, do której uczestnicy muszą ukończyć kurs. Po upływie terminu administrator otrzyma powiadomienie e-mail z listą osób, które nie ukończyły kursu.`
* EN: `Mark this course as mandatory for the selected group. You must set a completion deadline. After the deadline, the administrator will receive an email notification listing participants who have not completed the course.`
* 
### Issue 2: 

clicking on the input should open date picker:

<img width="900" height="943" alt="Image" src="https://github.com/user-attachments/assets/8cb77c04-5491-4a78-a646-d05f40329c35" />

I see that icon is clickable - that is good. Let's also trigger opening the date picker while clicking on input

### Issue 3: 

It is also not easy to see if group has that course mandatory -> admin needs to click on "Enroll group" button, then click on the checkbox for that group and only then the mandatoriness with the date is displayed. Let's change the behavior and if mandatory toggle is on, then always show it (along with the due date), so that admin does not have to click on the checkbox:

<img width="847" height="660" alt="Image" src="https://github.com/user-attachments/assets/0aea6b43-2347-490d-a051-9e0c9b2b99db" />

PR #1171, opened 2026-01-07
fix: improve mandatory course enrollment and group re-enrollment behavior

its-gabo:
## Issue(s)
- #1149 

## Overview 
- Improve mandatory course enrollment UI with tooltip copy and a date picker matching the registration flow, including locale and year dropdowns.
- Fix group enrollment behavior to only submit selected groups and allow re-enrolling users after group unenroll.

## Business Value
- Makes mandatory course settings clearer and easier to use for admins.
- Prevents accidental bulk enrollments and restores expected re-enrollment flow.

## Screenshot / Video
image
image

### DIFF
--- a/apps/web/app/modules/Admin/EditCourse/CourseEnrolled/GroupEnrollItem.tsx
+++ b/apps/web/app/modules/Admin/EditCourse/CourseEnrolled/GroupEnrollItem.tsx
@@ -4,8 +4,10 @@ import { enUS, pl } from "date-fns/locale";
 import { useMemo, useState } from "react";
 import { useFormContext, useWatch } from "react-hook-form";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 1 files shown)


apps/web/app/modules/Admin/EditCourse/CourseEnrolled/GroupEnrollItem.tsx
```diff
@@ -22,12 +32,25 @@ type Props = {
 };
 
 export const GroupEnrollItem = ({ index, id, name, usersCount, isGroupEnrolled }: Props) => {
-  const { t } = useTranslation();
+  const { t, i18n } = useTranslation();
```
Japrolol: You could potentially use our `useLanguageStore()` for language

Issue #1012, opened 2025-11-27
Fix broken true/false quiz


marektro:
There is an issue preventing a true/false answer to be correct. In the attached example the question is set to be correctly false, but answering neither true or false detects it as a right answer:
<img width="1914" height="1030" alt="Image" src="https://github.com/user-attachments/assets/4f5c65b9-f4cc-4cf1-8732-b282c577103f" />
<img width="1914" height="1035" alt="Image" src="https://github.com/user-attachments/assets/aa850bff-2d1f-4dc2-a318-c0ff8826a4c0" />

https://github.com/user-attachments/assets/1789ef8b-12a9-41ca-86ea-3e78d6866c7d

PR #1173, opened 2026-01-08
fix: show correct and incorrect true/false answers based on student responses

its-gabo:
# Issue(s)
- #1012 

## Overview
- Fix true/false quiz evaluation defaults to submit explicit boolean answers.
- Require true/false questions to be answered before submit.
- Render true/false correctness based on the student answer matching `question_answer_options.isCorrect`.

## Business Value
- Prevents false answers from being marked incorrect and avoids accepting unanswered true/false questions, improving quiz accuracy.

## Screenshots / Video

https://github.com/user-attachments/assets/a9234dd3-9f35-4c8e-975c-ac6f13909090

### DIFF
--- a/apps/web/app/modules/Courses/Lesson/Question/TrueOrFalse.tsx
+++ b/apps/web/app/modules/Courses/Lesson/Question/TrueOrFalse.tsx
@@ -26,61 +26,59 @@ export const TrueOrFalse = ({ question, isCompleted }: TrueOrFalseProps) => {
       questionType="trueOrFalseQuestion"
       questionNumber={question.displayOrder ?? 0}
     >
-      {question.options?.map(
-        ({ optionText, id, studentAnswer = false, isStudentAnswer }, index) => {
-          const isAnswered = studentAnswer === "true" || studentAnswer === "false";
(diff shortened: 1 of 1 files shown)


apps/web/app/modules/Courses/Lesson/Question/TrueOrFalse.tsx
```diff
@@ -27,57 +27,59 @@ export const TrueOrFalse = ({ question, isCompleted }: TrueOrFalseProps) => {
...
-                  })}
-                  {...(studentAnswer === "false" && { checked: true })}
-                  {...register(`trueOrFalseQuestions.${question.id}.${id}`)}
-                  name={`trueOrFalseQuestions.${question.id}.${id}`}
-                  type="radio"
-                  value="false"
-                />
-                {t("studentLessonView.other.false")}
-              </label>
+        ({ optionText, id, studentAnswer = false, isStudentAnswer }, index) => {
```
Japrolol:
```suggestion
        ({ optionText, id, studentAnswer, isStudentAnswer }, index) => {
```
studentAnswer will never be returned as boolean from the backend, so default value as boolean doesn't seem to make sense

PR #1184, opened 2026-01-09
feat: use SEO-friendly URLs for courses

mmienkoselleo:
## Issue(s)
[https://github.com/Selleo/mentingo/issues/1025](https://github.com/Selleo/mentingo/issues/1025)

## Overview 
Each course is assigned an SEO-friendly slug, which works in conjunction with the previous ID reference (the slug is not frozen, so I thought it was important to leave the resource reference method safer than the slug). Courses are listed by slug, so by default, every user will access links by slug. Editing/admin pages, etc. are not slugged because they are not part of SEO - however, there is nothing to prevent this from being extended in this way

In the case of a slug conflict, each subsequent slug will receive the suffix `-N`, where `N` is an integer one greater than the previous one (or `-1` by default for the first conflict) used in slugs with the same base

## Business Value
Sharing looks more serious and makes it easier to associate the link with the resource. Search engines and other external traffic will also treat it more naturally

## Screenshots / Video
image

### Notes 
To increase satisfaction with this feature, I suggest rethinking it. As I wrote above, they are not frozen, so they can be changed even after sharing. I have encountered two solutions, both equally good, but in different ways:
- hiding the short ID in the slug, e.g., `36c79-rest-of-the-slug`, where the text after the hash is not taken into account at all (this gives 36^5 = 60,466,176 possible combinations, I don't see any risk of someone creating so many courses at once)
- allowing the slug to be changed (or removed) only manually, showing the editor what the consequences will be
However, I believe that the feature is now fully competent - it depends very much on how someone uses it

### DIFF
--- a/apps/api/src/courses/course-slug.service.ts
+++ b/apps/api/src/courses/course-slug.service.ts
@@ -8,6 +8,8 @@ import { validate as uuidValidate } from "uuid";
 import { DatabasePg } from "src/common";
 import { courses, courseSlugs } from "src/storage/schema";
 
+import type { SupportedLanguages } from "@repo/shared";
+
 export type CourseSlugResult =
   | { type: "found"; courseId: string; slug: string }
   | { type: "redirect"; courseId: string; slug: string }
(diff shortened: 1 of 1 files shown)


PoslinskiNet:
It would be nice to have working link, even if the slug changes [so the old one will redirect with 300 to the new one - for SEO it might be a deal breaker and won't confuse end users.

its-gabo (approved):
Looks good 🍻

Rxbsxn (approved):

PR #1196, opened 2026-01-14
chore: remove Creative Commons licenses from approved options

k1eu:
License changes to not let CC licenses be used in the project


its-gabo:
@PoslinskiNet depends if @k1eu wants to add something to this PR

If not we can merge it right away

PR #1198, opened 2026-01-14
feat: allow admins to transfer course ownership

Japrolol:
## Issue(s)
[1169](https://github.com/Selleo/mentingo/issues/1169)

## Overview 

- Enables course ownership transfer, so admins can transfer courses between users, which is useful for e.g transferring ownership before deleting a user. It transfers all statistics to new owner.
- Fixed a flow, where user on logout and then login couldn't enter admin courses page, until the page was reloaded.

## Business Value

- Sometimes a course changes ownership due to a magnitude of reasons e.g user quits a company and will have it's account deleted but the course is valuable, so we can transfer the ownership to a different user.

## Screenshots / Video
https://github.com/user-attachments/assets/4fa15ec0-a760-43a4-8664-52d8d6456987


Japrolol:
Interesting thought. Maybe a scrollable select with a search bar would work better🤔

PR #1210, opened 2026-01-16
chore: remove vendor lock-in for video storage

its-gabo:
## Issue(s)
- #1204 

## Overview
- Add provider-based video uploads with Bunny/S3 selection and S3 multipart support.
- Update course/lesson content handling to flag missing Bunny embeds and avoid broken video embeds.
- Refactor file/course services for clearer upload flow and lesson decoration logic.
- Improve RichText video embed UI and i18n for tooltips/labels.

## Business Value
- Ensures video uploads continue working when Bunny is not configured by falling back to S3 multipart uploads.
- Improves admin content editing experience by surfacing embed errors and keeping UI accessible/localized.
- Reduces maintenance risk through clearer, more modular upload logic.

### Notes 
- [ ] Fired up e2e tests and got a positive result

### DIFF
--- a/apps/api/src/file/file.service.ts
+++ b/apps/api/src/file/file.service.ts
@@ -9,15 +9,7 @@ import {
   InternalServerErrorException,
   Logger,
 } from "@nestjs/common";
-import {
-  ALLOWED_EXCEL_FILE_TYPES,
-  ALLOWED_LESSON_IMAGE_FILE_TYPES,
-  ALLOWED_PDF_FILE_TYPES,
-  ALLOWED_PRESENTATION_FILE_TYPES,
-  ALLOWED_VIDEO_FILE_TYPES,
-  ALLOWED_WORD_FILE_TYPES,
-  VIDEO_UPLOAD_STATUS,
-} from "@repo/shared";
+import { ALLOWED_VIDEO_FILE_TYPES, VIDEO_UPLOAD_STATUS } from "@repo/shared";
 import { TypeCompiler } from "@sinclair/typebox/compiler";
 import { CacheManagerStore } from "cache-manager";
 import { parse } from "csv-parse";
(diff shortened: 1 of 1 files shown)


apps/api/src/file/file.service.ts
```diff
@@ -185,74 +211,174 @@ export class FileService {
...
+  private async initProviderUpload(
+    provider: VideoStorageProvider,
+    payload: VideoProviderInitPayload,
+    uploadId: string,
+    placeholderKey: string,
+  ) {
     try {
-      const response = await this.bunnyStreamService.createVideo(title || filename);
-      guid = response.guid;
+      return await provider.initVideoUpload(payload);
```
Japrolol:
```suggestion
      return provider.initVideoUpload(payload);
```

apps/api/src/file/guards/file.guard.ts
```diff
@@ -1,8 +1,38 @@
...
 
 import { EXTENSION_TO_MIME_TYPE_MAP } from "../file.constants";
 
+import type { FileTypeResult } from "file-type";
+
+type FromBufferFn = (buffer: Buffer) => Promise<FileTypeResult | undefined>;
+
+let fileTypeFromBufferFn: FromBufferFn | null = null;
+
+const getFileTypeFromBuffer = async (buffer: Buffer) => {
```
Japrolol: What's the reason for this complexity? Isn't this enough:
```ts
  static async getFileType(file: Express.Multer.File) {
    const { fileTypeFromBuffer } = await loadEsm<typeof import("file-type")>("file-type");

    return fileTypeFromBuffer(file.buffer);
  }
```

If there are no bytes it will return undefined by default

apps/web/app/modules/common/store/useVideoUploadResumeStore.ts
```diff
@@ -3,8 +3,9 @@ import { persist } from "zustand/middleware";
 
 export type StoredVideoUpload = {
   uploadId: string;
-  bunnyGuid: string;
+  bunnyGuid?: string;
   fileKey: string;
+  provider?: "bunny" | "s3";
```
Japrolol: Maybe use the shared VideoProviderType here?

Rxbsxn (approved):

PR #1229, opened 2026-01-22
chore: prepare database for multi-tenant support

its-gabo:
## Issue(s)
- #1227 

## Overview 
- Introduces full multi-tenant support across DB schema and data flows: creates `tenants` table, adds `tenant_id` FKs, enforces non-null constraints, and backfills legacy records.
- Adds RLS-aware scoped DB connection plus tenant-aware cron jobs, factories, helpers, seeds, and scripts for updating legacy tenant names/hosts.
- Improves performance and safety with new indexes on tenant-related tables and readonly DI for cron/auth services; updates user-related types to use `ActorUserType`.
- Fixes backend e2e setup to align with the new multi-tenant architecture.

## Business Value
- Ensures strict tenant data isolation and consistency, reducing risk of cross-tenant leakage.
- Speeds per-tenant queries and scheduled aggregations, keeping reporting performant at scale.
- Aligns authentication and user typing with multi-tenant rules, lowering maintenance overhead.

### DIFF
--- a/apps/api/src/auth/resend-verification-mail-cron.ts
+++ b/apps/api/src/auth/resend-verification-mail-cron.ts
@@ -7,7 +7,7 @@ import { TenantDbRunnerService } from "src/storage/db/tenant-db-runner.service";
 @Injectable()
 export class ResendVerificationMailCron {
   constructor(
-    private authService: AuthService,
+    private readonly authService: AuthService,
     private readonly tenantRunner: TenantDbRunnerService,
   ) {}
 
--- a/apps/api/src/courses/course.cron.ts
+++ b/apps/api/src/courses/course.cron.ts
@@ -8,7 +8,7 @@ import { CourseService } from "./course.service";
 @Injectable()
 export class CourseCron {
   constructor(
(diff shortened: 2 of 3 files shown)


apps/api/test/helpers/test-helpers.ts
```diff
@@ -36,24 +36,29 @@ export function signInAs(userId: string, jwtService: JwtService): string {
   return jwtService.sign({ sub: userId });
 }
 
-export async function truncateAllTables(connection: DatabasePg): Promise<void> {
+export async function truncateAllTables(
+  connection: DatabasePg,
+  scopedConnection: DatabasePg,
```
mikoscz: I would suggest removing 'scopedConnection' from `truncateAllTables`, per the docs: https://www.postgresql.org/docs/current/ddl-rowsecurity.html `Operations that apply to the whole table, such as TRUNCATE and REFERENCES, are not subject to row security.` This can be misleading to future users who might expect it to be scoped to the tenant. I suggest moving the settings insert to a separate helper.

PR #1251, opened 2026-01-30
feat: update course list with a modern layout

DawidHaratyk:
## Issue(s)
https://github.com/Selleo/mentingo/issues/882

## Overview 
Implemented modern course list

## Business Value
This makes our app better in the context of UI, it's more an app that everyone wants to use as it just looks better and more encouraging for users

## Screenshots / Video

https://github.com/user-attachments/assets/bd6d6044-800b-426b-8691-ab4ab000a844

image

### Notes 

TBD:
- there are still some issues with running trailer eg. controls are visible for a moment when starting a video, needs to customly adjust the bunny provider propably

My suggestions that you can review and then we can implement:
- Carousel button arrows can be in the middle of the cards, not on top
- Course cards can overflow carousel buttons


apps/web/app/modules/Courses/components/modern/CoursesHeader.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { Button } from "~/components/ui/button";
+
+const CoursesHeader = () => {
+  return (
+    <header className="sticky top-0 w-full flex justify-end p-2 -mt-16 z-[40]">
+      <div className="items-center gap-3 flex">
+        <Link to="/admin/courses">
+          <Button variant="outline" className="w-full">
+            Manage Courses
```
Rxbsxn: Is translation needed here?

PR #1267, opened 2026-02-05
chore: improve modern courses view 

DawidHaratyk:
## Issue(s)
https://github.com/Selleo/mentingo/issues/1257
https://github.com/Selleo/mentingo/issues/1264

## Overview 
As in tasks +
- Fixed a case where user is enrolled via group that is currently enrolled individually, basically it was working but course due date was not updated properly and in result student in the group couldn't see the course due date

## Business Value
- More enjoyable UI
- Less users being irritated using courses carousel
- Elements are more visible for users (changed hero contrast)
- Users can use the courses list on big resolutions

## Screenshots / Video
Fixed layout when resolution is high

image

Less padding space between carousel sections

image

Video

https://github.com/user-attachments/assets/a0c56144-b11c-42be-afa2-3cb149255297

### DIFF
--- a/apps/api/src/courses/__tests__/course.controller.e2e-spec.ts
+++ b/apps/api/src/courses/__tests__/course.controller.e2e-spec.ts
@@ -2045,7 +2045,7 @@ describe("CourseController (e2e)", () => {
         });
 
         describe("when group has users", () => {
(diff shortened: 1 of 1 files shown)


DawidHaratyk:
> Build 🔴 image

I've fixed the issue, it should work right now

PR #1276, opened 2026-02-06
feat: introduce super-admins role with ability to manage tenants

its-gabo:
## Issue(s)
- #1227 

## Overview
- Implement super-admin with tenant edition.
- Refactor backend and frontend.
- Centralize seed-related helpers, ensure default global settings, and refine user invite notification logic.
- Implement dedicated inactive tenant page with specific API route allowances and updated error handling.

## Business Value
- Improves tenant administration and lifecycle handling, enabling super-admin workflows and clearer inactive-tenant UX.

### Notes
- [ ] Fired up e2e tests and got a positive result

### DIFF
--- a/apps/web/e2e/tests/admin-student/assigning-student-to-course.spec.ts
+++ b/apps/web/e2e/tests/admin-student/assigning-student-to-course.spec.ts
@@ -222,28 +222,23 @@ const clickLessonWithRetry = async (page: Page, chapterButton: Locator, lessonLi
   await expect(page).toHaveURL(/course\/.*\/lesson\/.*/);
 };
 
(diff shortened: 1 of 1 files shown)


apps/reverse-proxy/Caddyfile
```diff
@@ -8,6 +8,16 @@ app.lms.localhost {
     }
 }
 
+tenant1.lms.localhost tenant2.lms.localhost tenant3.lms.localhost tenant4.lms.localhost {
```
Rxbsxn: Cannot it be something like `*.lms.localhost`? This should be quite flexible when it comes to the namings

Rxbsxn (approved):

PR #1290, opened 2026-02-11
fix: eliminate flickering in fill-in-the-blanks questions

DawidHaratyk:
## Issue(s)
https://github.com/Selleo/mentingo/issues/1284
https://github.com/Selleo/mentingo/issues/1277

## Overview 
As in tasks

## Business Value
User is able to see text in proper place


its-gabo:
@DawidHaratyk The issue with the bouncing still happens

https://github.com/user-attachments/assets/8fbe4241-6618-4796-8a40-e1a82c487f9c

PR #1319, opened 2026-02-23
feat: introduce AI-powered course generation

Japrolol:
## Issue(s)
[1313](https://github.com/Selleo/mentingo/issues/1313)

## Overview 

- Implemented integration with external luma service
- Added generate with AI drawer and button
- 

## Business Value

- It allows content creators to create courses more easily using the AI tool.

## Screenshots / Video
https://github.com/user-attachments/assets/16ba77b3-9b75-42ff-add6-bf5552104f98

### DIFF
--- a/apps/api/src/ingestion/services/ingestion.service.ts
+++ b/apps/api/src/ingestion/services/ingestion.service.ts
@@ -37,13 +37,13 @@ export class IngestionService {
       throw new BadRequestException("Exceeded max number of files");
     }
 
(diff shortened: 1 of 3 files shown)


its-gabo:
@Japrolol  Works well except 2 things we talked about. 

It can take ages until it starts to generate the course. It could be frustrating to end users. And as you saw it was a struggle for me to get it to generate the course.

Second things is that messages order got messed up and got me confused.

Also as you said it could use a bit of prompt enhancement, but the generated course itself is looking great

PR #1428, opened 2026-04-09
chore: update storage paths to make file management simpler and more organized

its-gabo:
## Issue(s)
- #1427 

## Overview
Updated file upload handling so uploaded objects are stored under tenant-prefixed S3 keys. This change threads `tenantId` through the main upload paths, tenant-aware video initialization, thumbnail generation, certificate share image storage, and the remaining direct upload call sites.

## Business Value
This keeps uploaded assets isolated by tenant at the storage layer, which improves data organization, reduces collision risk, and makes tenant-level cleanup and migration work safer.

### DIFF
--- a/apps/api/src/file/file.service.ts
+++ b/apps/api/src/file/file.service.ts
@@ -128,7 +128,7 @@ export class FileService {
       throw new BadRequestException("File upload failed - empty file");
     }
 
-    if (!tenantId) throw new BadRequestException("Missing tenant context for file upload");
+    if (!tenantId) throw new BadRequestException("files.toast.missingTenantContext");
 
     const isVideo = file.mimetype.startsWith("video/");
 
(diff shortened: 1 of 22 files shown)


apps/api/src/file/providers/s3-video.provider.ts
```diff
@@ -22,9 +24,12 @@ export class S3VideoProvider implements VideoStorageProvider {
   }
 
   async initVideoUpload(payload: VideoProviderInitPayload): Promise<VideoProviderInitResult> {
-    const { filename, mimeType, resource } = payload;
+    const { filename, mimeType, resource, tenantId } = payload;
+
+    if (!tenantId) throw new BadRequestException("Missing tenant context for video upload");
```
Japrolol: Can we return translation?

apps/api/src/file/file.service.ts
```diff
@@ -122,11 +123,13 @@ export class FileService {
     }
   }
 
-  async uploadFile(file: Express.Multer.File, resource: string) {
+  async uploadFile(file: Express.Multer.File, resource: string, tenantId?: UUIDType) {
     if (file.size === 0) {
       throw new BadRequestException("File upload failed - empty file");
     }
 
+    if (!tenantId) throw new BadRequestException("Missing tenant context for file upload");
```
Japrolol: Same as above

Japrolol (approved):

Issue #1354, opened 2026-03-11
Make E2E tests work again


maciejhubert:
We need to to a few things regarding e2e:
* fix 2 flaky tests
* fix remaining 7 failing tests
* add test to check video on Firefox (and add Firefox to other tests)

PR #1472, opened 2026-04-29
test: extend coverage with settings

its-gabo:
## Issue(s)
- #1354 

## Overview
This change significantly extends end-to-end coverage for the **Settings** area across account, organization, platform preferences, navigation access, permissions visibility, support mode, and related admin flows.

It adds a broad set of new Playwright specs and shared test handles for settings, plus small supporting updates in settings components and utility code to improve testability and flow reliability.

## Business Value
This reduces regression risk in one of the most configuration-heavy parts of the product and increases confidence in releases by validating key settings workflows from the end-user perspective.

The added coverage should also shorten debugging and QA cycles for settings-related changes.


apps/web/app/modules/Dashboard/Settings/components/SettingsNavigationTabs.tsx
```diff
@@ -60,11 +61,21 @@ export function SettingsNavigationTabs({
                 <TabsTrigger
                   key={tab.value}
                   value={tab.value}
+                  data-testid={
+                    tab.value === SETTINGS_TABS.ACCOUNT
```
Japrolol: Can we do this with some switch statement instead of double ternary?

Japrolol (approved):

PR #1506, opened 2026-05-14
chore: improve fill-in-the-blank exercises with support for repeated words

its-gabo:
## Overview

Improves fill-in-the-blanks quiz handling by tying each blank placeholder to a concrete answer option id instead of relying only on positional `[word]` markers.

Main changes:

- Added `<blank-answer-{answerId}>` markers for fill-in-the-blanks text and drag-and-drop questions.
- Added a migration to backfill existing localized question descriptions from legacy `[word]` markers to answer-id markers.
- Updated quiz answer serialization/evaluation so submitted fill-in-the-blanks answers can be stored and read by answer id while keeping legacy positional fallback support.
- Updated student fill-in-the-blanks text and drag-and-drop rendering to map blanks, submitted answers, and completed feedback by answer id.
- Allowed duplicate words to be used as separate blank answers in the admin quiz editor and student drag-and-drop flow.
- Updated SCORM export runtime fill-in-the-blanks rendering and correctness checks to understand answer-id markers.
- Improved quiz submit validation errors to report missing question names.
- Made certificate creation idempotent for an already completed course and covered retaking a final quiz after certificate generation.

## Business Value

This makes fill-in-the-blanks quizzes more reliable for authors and students. Authors can now create questions where the same word appears multiple times, and the platform can still distinguish which blank each answer belongs to.

For learners, quiz feedback and retakes are more consistent because submitted answers are restored against the intended blanks instead of depending on display order or repeated word values. Existing courses are migrated forward, and SCORM exports keep the same behavior.

## Screenshots / Video

https://github.com/user-attachments/assets/00f25f2d-a8cf-40e8-8963-57ac4265ee77


apps/api/src/questions/question.service.ts
```diff
@@ -20,6 +20,11 @@ import type {
 } from "src/lesson/lesson.schema";
 import type * as schema from "src/storage/schema";
 
+type FormattedAnswer = {
```
Japrolol: Put this in a separate file

Japrolol (approved):

Issue #1512, opened 2026-05-15
Add S3 timeout and stronger error handling


its-gabo:
S3/file delivery can leave requests hanging and keep DB transactions open. Add a configurable timeout around S3 operations and make errors fail fast with clear handling/logging.

PR #1516, opened 2026-05-15
chore: improve file storage reliability with better error handling and timeouts

its-gabo:
## Issue(s)
- #1512 

## Overview

Enhances S3 and file delivery error handling so storage operations fail predictably and file streams are cleaned up when responses fail, close early, or stall.

- Adds timeout-aware S3 SDK calls using `AbortController` and a shared 5-minute operation timeout.
- Logs S3 failures with operation, key, status code when available, and original stack traces.
- Propagates the S3 timeout to file streaming responses and destroys streams on response errors, early closes, and stream timeouts.
- Improves file service logging for upload, delete, and retrieve failures while preserving API-facing exceptions.
- Adds unit coverage for S3 operation timeouts and file stream cleanup on response failure or close.

## Business Value

This reduces the risk of API requests hanging indefinitely when S3 or MinIO is slow or unavailable, prevents leaked file streams during interrupted downloads, and gives maintainers clearer logs for diagnosing storage-related incidents. End users should see more consistent failure behavior instead of stalled uploads or downloads.

### DIFF
--- a/apps/api/src/file/file.service.ts
+++ b/apps/api/src/file/file.service.ts
@@ -1,4 +1,5 @@
 import { randomUUID } from "crypto";
+import path from "node:path";
 import { Readable } from "stream";
 
 import {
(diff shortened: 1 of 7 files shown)


Japrolol (approved):

apps/api/src/file/file.service.ts
```diff
@@ -152,33 +152,30 @@ export class FileService {
...
-    try {
-      if (isVideo) {
-        throw new BadRequestException("Video uploads must use the TUS endpoints");
-      }
+    if (isVideo) {
+      throw new BadRequestException("Video uploads must use the TUS endpoints");
+    }
 
-      const fileExtension = file.originalname.split(".").pop();
+    const fileExtension = file.originalname.split(".").pop();
```
mikoscz: Consider `path.extname`

Issue #1521, opened 2026-05-18
Remove autoplay


maciejhubert:
Let's remove autoplay feature. It causes more problems than we anticipated.

Possibly in the future we will have video-based courses, who knows, and then it would make sense. But for now let's remove that feature.

PR #1524, opened 2026-05-18
chore: remove autoplay functionality

its-gabo:
## Issue(s)

- #1521 

## Overview

Removes the global video autoplay flow and simplifies rich-text video playback to an inline, user-activated player.

The change removes autoplay, countdown, play-next, fullscreen-preservation, and singleton video-player state while preserving content lesson completion when the last video in a lesson ends. It also adds playback speed control to the Video.js control bar with rates from `0.5x` to `2x`, including support for the custom Vimeo tech.

## Business Value

This makes video playback more predictable for students and removes automatic behavior that can unexpectedly start or advance videos. Students still get lesson progress completion after finishing the final video, and they gain playback speed controls for more flexible learning.

## Screenshots / Video

https://github.com/user-attachments/assets/98a3e951-d986-41bf-9652-24223cc41a16


apps/web/app/components/RichText/Viever.tsx
```diff
@@ -3,37 +3,38 @@ import { useCallback, useEffect, useMemo, useRef } from "react";
...
   style?: "default" | "prose";
   className?: string;
-  variant?: "default" | "article" | "news" | "content";
-  onVideoEnded?: VideoEndedHandler;
-  videoAutoplayPolicy?: RichTextVideoAutoplayPolicy;
+  variant?: RichTextViewerVariant;
+  onVideoEnded?: (index: number | null) => void;
 };
 
+export const RICH_TEXT_VIEWER_VARIANT = {
```
Japrolol: Do we want to keep this enum in here, or can we extract it to separate file?

apps/web/app/components/VideoPlayer/Video.tsx
```diff
@@ -1,120 +1,47 @@
...
-  index,
-}: Props) {
-  const ref = useRef<HTMLDivElement>(null);
-  const [hasAutoTriggered, setHasAutoTriggered] = useState(false);
-  const { setVideo, setPlaceholderElement, registerVideoActivator, state } = useVideoPlayer();
-
-  const { autoplay: autoplayEnabled, setAutoplaySettings } = useVideoPreferencesStore();
+export function Video({ src, url, provider, index = null, onEnded }: Props) {
+  const [isActive, setIsActive] = useState(false);
+  const [aspectRatio, setAspectRatio] = useState("16 / 9");
```
Japrolol: This feels brittle, is there a different approach?

I understand that this is the simplest way as we just pass it to styles, but just a thing to keep in mind.

Japrolol (approved):

apps/web/app/components/VideoPlayer/Video.tsx
its-gabo: The issue here is that when we just simply put `aspect-video` on the element then we have black bars on both sides of the player. Also if the video is let's say 16:10 then we have the same issue.

It looks like this:
image

PR #1537, opened 2026-05-20
feat: allow courses to repeat in cycles for continuous learning programs

its-gabo:
## Issue(s)
- #1522 

## Overview

Adds cyclic course certificate support for courses that require certificates to expire and be earned again.

The change introduces certificate validity settings on course certificates, including period-based validity and fixed expiration dates. Admins can preview the impact of validity changes, choose whether to apply new validity to existing active certificates, and manually reset certificates for all users, selected groups, or selected users.

On the backend, certificates now have an active/archive lifecycle with issued, expiry, archive, archive reason, and expiration-warning timestamps. Expired or manually reset certificates are archived, related course progress is reset, activity logs are recorded, and certificate expiration or reset emails are published through the outbox. Daily tenant-scoped cron jobs send upcoming expiration warnings and expire certificates at midnight.

## Business Value

Course managers can configure certificates that remain valid only for a controlled period. This supports recurring compliance, certification, and retraining flows where learners must retake a course after their certificate expires.

Admins also get safer operational controls: they can see how many certificates will be affected before changing validity, reset only the intended audience, optionally notify learners by email, and keep an audit trail for manual and system-driven certificate changes.

## Screenshots / Video

https://github.com/user-attachments/assets/9c7e25df-7a90-4d77-9d54-9b1e0fa17471


apps/api/src/certificates/certificates.controller.ts
```diff
@@ -138,6 +154,84 @@ export class CertificatesController {
...
+  async getCertificateResetUsers(
+    @Param("courseId") courseId: UUIDType,
+    @Query("page") page: number | undefined,
+    @Query("perPage") perPage: number | undefined,
+    @Query("search") search: string | undefined,
+    @CurrentUser() currentUser: CurrentUserType,
+  ): Promise<PaginatedResponse<CertificateResetUsersResponse>> {
+    const data = await this.certificatesService.getCertificateResetUsers(
+      courseId,
+      { page, perPage, search } satisfies CertificateResetUsersQuery,
```
Japrolol: Satisfies is redundant here

apps/api/src/certificates/certificates.service.ts
```diff
@@ -151,6 +219,191 @@ export class CertificatesService implements OnModuleDestroy {
...
+
+    await this.db.transaction(async (trx) => {
+      await this.archiveAndResetCertificates(
+        certificatesToReset,
+        CERTIFICATE_ARCHIVE_REASONS.MANUAL_RESET,
+        trx,
+      );
+    });
+
+    if (body.sendEmail ?? true) {
```
Japrolol: Cleaner approach would be:

```suggestion
    if (body.sendEmail !== false) {
```

apps/api/src/certificates/certificate.repository.ts
```diff
@@ -246,56 +296,513 @@ export class CertificateRepository {
...
+
+  async updateActiveCertificateExpirations(
+    updates: { certificateId: UUIDType; expiresAt: Date | null }[],
+    trx?: DatabasePg,
+  ) {
+    if (!updates.length) return [];
+
+    const dbInstance = trx || this.db;
+
+    return processInBatches(
```
Japrolol: This still executes one UPDATE per certificate. `processInBatches`
  limits concurrency, but it does not reduce query count, so applying
  validity to many certificates scales as N update queries. Can we
  make this set-based instead, e.g. one UPDATE using SQL date
  arithmetic for relative validity, or a single UPDATE ... FROM/
  VALUES/CASE keyed by certificate id if these app-computed values
  need to stay?

apps/api/src/certificates/certificates.service.ts
```diff
@@ -151,6 +219,191 @@ export class CertificatesService implements OnModuleDestroy {
...
+
+    const updates = certificatesToUpdate.map((certificate) => ({
+      certificateId: certificate.id,
+      expiresAt: this.calculateCertificateExpiry(
+        certificateValidity,
+        new Date(certificate.issuedAt),
+      ),
+    }));
+
+    const immediatelyExpiredCertificates = certificatesToUpdate.filter((certificate) => {
```
Japrolol: This is O(N²): for every certificate we scan `updates` again with
  `.find`. Since we prefer map-based M+N patterns, can we either
  compute `updates` and `immediatelyExpiredCertificates` in one pass,
  or build an `updatesByCertificateId` map before filtering?

apps/web/app/modules/Admin/EditCourse/CourseSettings/components/useCertificateValiditySettings.ts
Japrolol, line 1: Can we extract reducer to separate file?

apps/web/app/modules/Admin/EditCourse/CourseSettings/components/CourseCertificateSetting.tsx
```diff
@@ -143,75 +174,102 @@ const CourseCertificateSetting = ({
...
-              <Icon name="TrashIcon" className="mr-2" />
-              {t("adminCourseView.settings.button.removeCertificateSignature")}
+              <Icon name="Eye" className="mr-2 size-4" />
+              {t("adminCourseView.settings.other.certificatePreviewButton")}
             </Button>
-          )}
+          </div>
+
+          <div
+            className={`w-full max-w-xl space-y-2 transition-opacity ${
```
Japrolol: Can we use cn?

apps/web/app/modules/Admin/EditCourse/CourseSettings/components/CertificateValiditySection.tsx
```diff
@@ -0,0 +1,105 @@
...
+  onSave,
+}: CertificateValiditySectionProps) {
+  const { t } = useTranslation();
+
+  const isSaveButtonDisabled =
+    disabled || !hasValidityChanges || isCheckingValidityImpact || isUpdatingCourseSettings;
+
+  return (
+    <div
+      className={`flex flex-col gap-4 transition-opacity ${
```
Japrolol: Use cn for merged classnames

Japrolol (approved):
Mostly looks good, small feedback. Just check also whether I missed in review other places where you use normal ternaries inside className string template instead of `cn()`

Japrolol (approved):

Issue #1433, opened 2026-04-14
Allow translating various resources


maciejhubert:
Let's allow translating:

- Categories
- Groups

Let's also allow translating announcements - if an announcement is not available in the language of a user, do not show i (the same as in news, articles, Q&A)

PR #1539, opened 2026-05-21
feat: add translation support for course categories

its-gabo:
## Issue(s)
- #1433 

## Overview

Adds translation support for categories across the API and admin web UI.

- Stores category titles as localized JSONB values with `baseLanguage` and `availableLocales`.
- Adds API support for listing and reading categories in a requested language.
- Adds category language management endpoints for creating a language variant, deleting a language variant, and changing the base language.
- Updates category create/update flows to include language information and localized uniqueness checks.
- Regenerates Swagger and web API client artifacts for the new category contract.
- Updates the admin category pages with language selection, language creation/deletion dialogs, base language controls, and localized category queries.
- Adds and updates API and E2E test coverage/factories for translated categories.

## Business Value

Admins can manage category names in multiple supported languages, keeping category labels aligned with the learner's selected language. This improves localization coverage in course discovery and admin workflows while preserving tenant-level uniqueness rules for each category's base title.

## Screenshots / Video

https://github.com/user-attachments/assets/523f4f3c-10f9-42dd-b6bb-8c839c4114a8


apps/api/src/category/category.service.ts
```diff
@@ -87,147 +95,275 @@ export class CategoryService {
...
+      .select({
+        ...getTableColumns(categories),
+        title: this.getLocalizedCategoryTitle(language),
+      })
       .from(categories)
-      .where(and(eq(categories.id, id)));
+      .where(eq(categories.id, id));
+
+    if (!category) {
+      throw new NotFoundException({ message: "adminCategoryView.error.categoryNotFound" });
```
Japrolol, line 115: Isn't the message by default returned as `{message: ...}`?

apps/web/app/modules/Admin/Categories/components/CategoryLanguagesSelector.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+import { CATEGORY_PAGE_HANDLES } from "../../../../../e2e/data/categories/handles";
+
+import type { IconName } from "~/types/shared";
+
+const languageOptions: {
+  key: SupportedLanguages;
+  iconName: IconName;
+  translationKey: string;
+}[] = [
```
Japrolol: This is used in every language selector. It could be useful to extract this to a separate file and re-use the same one everywhere.

apps/web/app/modules/Admin/Categories/components/CategoryLanguagesSelector.tsx
Japrolol, line 1: Could we maybe extract the language selector to a component and just pass callbacks to it, so we can re-use the same UI, while keeping logic different. At this point we have multiple language selectors? Don't we have a global LanguageSelector?

Japrolol (approved):

apps/web/app/modules/Admin/Categories/components/CategoryLanguagesSelector.tsx
its-gabo: Same thought crossed my mind, but i was unsure if there will be no case-specific things. 

I've done it in another PR with groups internationalization.

apps/web/app/modules/Admin/Categories/components/CategoryLanguagesSelector.tsx
its-gabo: Same as in the language selector. It is done in PR with groups internationalization

apps/api/src/category/category.service.ts
its-gabo: It is, but it's the same in both cases, whetever with {message: ''} or without it 😭

Issue #1527, opened 2026-05-18
Apply SCORM feedback, part 2


maciejhubert:
This scenario considers [this course](https://learn.selleo.com/course/8c6h8-warsztaty-modele-prawdopodobienstwa-w-grach-i-decyzjach) to be exported to SCORM and `imported as a SCORM lesson`.

### Issue 4
Remove "standard" label as it might be confusing what that means. It is enough to display SCORM label for SCORM courses:

<img width="1094" height="714" alt="Image" src="https://github.com/user-attachments/assets/ec06a09e-a340-403f-9c8d-f02bad88b747" />

### Issue 5
It is not possible to complete a SCORM lesson (note that the one I have imported contains several quizzes in separate SCOs).
Moreover, after completing a quiz, moving to the next SCO and then to previous SCO, results in quiz answers to be cleared out. Those two issues are probably related.

### Issue 6
Let's change this column name to `SCORM` and values should be displayed as check or no check.
<img width="1635" height="942" alt="Image" src="https://github.com/user-attachments/assets/94481bfd-3a6a-4cf7-83ca-6434067f478b" />

Issue #1526, opened 2026-05-18
Apply SCORM feedback, part 1


maciejhubert:
This scenario considers [this course](https://learn.selleo.com/course/8c6h8-warsztaty-modele-prawdopodobienstwa-w-grach-i-decyzjach) to be exported to SCORM and `imported as a SCORM course`.

### Issue 1
it is not possible to fail a quiz as far as I see:

<img width="779" height="627" alt="Image" src="https://github.com/user-attachments/assets/eac7a595-4a83-4c7d-9e69-2583909d7057" />

Passing threshold should work as established in the course.

### Issue 2
If possible, it would be good to retake a quiz. Not sure if we can limit the retakes count?

<img width="1130" height="726" alt="Image" src="https://github.com/user-attachments/assets/66f16d30-72e0-4fb4-944f-eb9c6c869c72" />

### Issue 3 (IGNORE, NOT AN ISSUE)
Let's hide language component in course edit mode as it does not make sense at the moment:

<img width="1512" height="866" alt="Image" src="https://github.com/user-attachments/assets/02cd3bc6-5a5b-457c-bf3c-1bb4236c1b4d" />

Japrolol:
Language component in my opinion makes sense because you might want to have different course titles per language

PR #1551, opened 2026-05-27
chore: improve the SCORM experience with usability and reliability enhancements

its-gabo:
## Issue(s)

- #1526 
- #1527 
- #1498 

## Overview

Improves SCORM import and runtime handling across API, web, shared constants, and the exported SCORM player.

- Moves SCORM package processing into a BullMQ worker so course creation, lesson creation, and lesson package attachment can return after staging the upload.
- Adds SCORM import socket events and shared action/message constants so the web app can show global processing, success, and failure toasts.
- Stages uploaded packages in S3, processes them asynchronously, marks packages ready when extraction succeeds, and cleans up staged files or failed entities when imports fail.
- Updates course and lesson queries after SCORM import status changes so admins see fresh course/lesson data without manual refresh.
- Handles missing SCORM courses more gracefully by redirecting users back to course lists instead of the generic home page.
- Prevents SCORM section navigation while runtime progress is saving and keeps launch runtime cache in sync after commit/finish.
- Adds a retake action to the exported SCORM quiz runtime and adjusts quiz action styling.
- Simplifies SCORM labeling in course admin views so only SCORM courses show the SCORM badge/column marker.

## Business Value

Admins can upload SCORM packages without waiting for all package extraction and file uploads to finish in the request cycle. This makes large SCORM imports more reliable, gives clear feedback while processing continues, and automatically removes failed imports so users are not left with broken courses or lessons.

Students get more reliable SCORM progress behavior, especially for multi-SCO packages, and exported quiz content can be retaken after submission.

### DIFF
--- a/apps/api/src/scorm/scorm-import.worker.ts
+++ b/apps/api/src/scorm/scorm-import.worker.ts
@@ -1,0 +1,76 @@
+import {
+  HttpException,
+  Injectable,
+  InternalServerErrorException,
+  type OnModuleDestroy,
+} from "@nestjs/common";
+import { Worker } from "bullmq";
+
+import { QUEUE_NAMES, QueueService } from "src/queue";
+import { TenantDbRunnerService } from "src/storage/db/tenant-db-runner.service";
+
+import { SCORM_IMPORT_JOB_NAME } from "./scorm-queue.service";
+import { ScormService } from "./scorm.service";
+
+import type {
+  ScormImportJobData,
+  ScormImportJobFailure,
+  ScormImportJobResult,
+} from "./scorm.types";
+import type { Job } from "bullmq";
+
+@Injectable()
+export class ScormImportWorker implements OnModuleDestroy {
+  private readonly worker: Worker<ScormImportJobData, ScormImportJobResult>;
+
+  constructor(
+    private readonly queueService: QueueService,
+    private readonly scormService: ScormService,
+    private readonly tenantDbRunnerService: TenantDbRunnerService,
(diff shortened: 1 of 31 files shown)


apps/api/src/scorm/scorm-import.worker.ts
```diff
@@ -0,0 +1,76 @@
...
+      {
+        connection: this.queueService.getConnection(),
+        concurrency: Number(process.env.SCORM_WORKER_CONCURRENCY || 1),
+      },
+    );
+  }
+
+  private async handleScormImport(job: Job<ScormImportJobData>): Promise<ScormImportJobResult> {
+    if (job.name !== SCORM_IMPORT_JOB_NAME) {
+      throw new InternalServerErrorException("adminScorm.errors.unknownImportJob");
```
Japrolol: Since this is an internal exception, maybe let's use a descriptive error message, rather than translation key?

apps/web/app/modules/Courses/CourseView/CourseView.page.tsx
```diff
@@ -64,10 +65,18 @@ export const clientLoader = async ({
   const url = new URL(request.url);
   const language = resolvePreferredLanguage(url);
 
-  const lookupResponse = await ApiClient.api.courseControllerLookupCourse({
```
Japrolol, line 67: Do we want to coll lookupResponse directly instead of through a mutation or query?

Japrolol (approved):

Issue #1547, opened 2026-05-26
Fix presentation viewer


Japrolol:
<img width="1512" height="784" alt="Image" src="https://github.com/user-attachments/assets/a4fd7237-828c-42f2-82f0-fe5e98cae16e" />

PR #1552, opened 2026-05-28
fix: restore presentation previews for uploaded files

its-gabo:
## Issue(s)
- #1547 

## Overview

Fixes presentation files not displaying properly by serving uploaded presentations through a generated PDF preview and rendering them with the existing PDF preview experience.

- Adds `preview=pdf` support to lesson, article, and news resource endpoints.
- Converts presentation files to PDF on demand with LibreOffice, stores the generated preview next to the original file in S3, and reuses it for later requests.
- Keeps the original file streaming behavior unchanged when no preview is requested.
- Installs LibreOffice and additional fonts in the API Docker image so conversion works in the deployed runtime.
- Replaces the local presentation document viewer with the shared PDF preview viewer.
- Splits the rich text PDF preview implementation into smaller viewer, reducer, constants, types, and utility modules.
- Adds localized presentation/PDF preview labels across supported locales.

## Business Value

Users can reliably view uploaded presentation resources inline instead of seeing broken or inconsistent previews. This improves lesson, article, and news content consumption for students and reduces the need to download files manually just to inspect presentation content.

## Screenshots / Video

https://github.com/user-attachments/assets/f1e4ccd8-deb0-4c39-9d81-7cf0fc253931


apps/api/src/file/file.service.ts
```diff
@@ -435,6 +444,87 @@ export class FileService {
...
+    if (await this.s3Service.getFileExists(pdfPreviewKey)) return pdfPreviewKey;
+
+    const tempDirectory = await mkdtemp(path.join(tmpdir(), "mentingo-presentation-"));
+    const inputExtension = path.extname(fileKey) || ".pptx";
+
+    const inputPath = path.join(tempDirectory, `presentation${inputExtension}`);
+    const outputPath = path.join(tempDirectory, "presentation.pdf");
+
+    try {
+      const fileBuffer = await this.s3Service.getFileBuffer(fileKey);
```
Japrolol, line 532: The preview generation reads the entire presentation into memory with s3Service.getFileBuffer(fileKey). Uploaded files can be up to MAX_FILE_SIZE = 1GB, so one preview request can allocate a huge
     buffer, then LibreOffice allocates more during conversion. This can OOM or stall the API process. Use a bounded size check before conversion, or stream to the temp file instead of buffering the
     whole object.

its-gabo: I've added a limit of 150MB for presentation uploads, because let's be honest that presentations will not have more than that.

Japrolol: Sounds good, how do images in presentations affect the size? As far as PDF's go, they can grow drastically in size once a few images appear. What about PPTX or others?

Issue #1505, opened 2026-05-13
Add course due dates notifications


maciejhubert:
Let's add a notification for due dates in mandatory courses.

So if a student is assigned in a group that has the courses mandatory, each such student should receive an email notification 7 days and 1 day before due date.

Also, if #1447 is done, let's add those due days to students' calendars.

maciejhubert:
You can do this part in a separate ticket;

> Also, if https://github.com/Selleo/mentingo/issues/1447 is done, let's add those due days to students' calendars.

PR #1567, opened 2026-06-02
feat: add course due date notifications

its-gabo:
## Issue(s)

[#1505](https://github.com/Selleo/mentingo/issues/1505)

## Overview

Adds course due date notifications for mandatory group course enrollments.

This PR creates and synchronizes calendar events for mandatory group course due dates, backfills existing due dates into calendar events, and exposes the new `course_due_date` calendar source through the API and generated web client. It also sends daily reminder emails and in-app announcements to students 7 days and 1 day before an incomplete mandatory course is due.

On the web side, course due date calendar events get their own styling and details dialog content, including due date, course, group, and a link back to the course. Related group enrollment mutations now invalidate the broader groups-by-course query key so due date metadata refreshes correctly.

## Business Value

Students get clearer visibility into required course deadlines directly in their calendar and receive timely reminders before those deadlines pass.

Admins and course managers can rely on mandatory group enrollment due dates being reflected consistently in calendar views and reminder communication, reducing missed deadlines and manual follow-up.

## Screenshots / Video

https://github.com/user-attachments/assets/2b1e1d9e-4f1a-4794-a86f-2b66d59c60f0


apps/api/src/calendar/calendar.service.ts
```diff
@@ -154,14 +191,22 @@ export class CalendarService {
 
     const listItem = this.mapLiveTrainingListItem(row, hosts, linkedCourses, currentUser);
     const isPrivilegedViewer = this.isPrivilegedViewer(row, hosts, currentUser);
+    const liveTrainingPayload = row.payload as LiveTrainingCalendarEventPayload;
+    const liveTrainingListPayload = (
+      listItem.payload as {
```
Japrolol: Can we introduce a named type for this payload shape instead of keeping the cast inline? This section is doing enough mapping already, and a small `LiveTrainingDetailsPayload`/similar type would make it easier to scan.

apps/api/src/calendar/calendar.service.ts
```diff
@@ -154,14 +191,22 @@ export class CalendarService {
...
+        liveTraining: LiveTrainingCalendarEventPayload["liveTraining"] & {
```
Japrolol: Same here. The inline intersection makes the payload mapping harder to read. A named type or small mapper would keep this maintainable.

apps/api/src/calendar/calendar.repository.ts
```diff
@@ -53,6 +62,92 @@ export class CalendarRepository {
...
+  ): Promise<GroupCourseDueDateRow[]> {
+    if (!groupIds.length) return [];
+
+    return this.db
+      .select({
+        courseId: groupCourses.courseId,
+        groupId: groupCourses.groupId,
+        dueDate: sql<string>`${groupCourses.dueDate}`,
+        calendarEventId: groupCourses.calendarEventId,
+        courseTitle: sql<LocalizedText>`${courses.title}`,
```
Japrolol, line 92: Do we need the full localized title object here because this row is used to create/sync calendar events? If yes, this is fine, but I would make that explicit in the type/helper so it does not look like a missing `getLocalizedSqlField` call.

apps/api/src/calendar/course-due-date-calendar.service.ts
```diff
@@ -0,0 +1,96 @@
...
+    const removableCalendarEventIds = calendarEventIds.filter((id): id is UUIDType => Boolean(id));
+
+    await this.calendarRepository.removeCourseDueDateCalendarEvents(removableCalendarEventIds);
+  }
+
+  private async syncGroupCourseDueDate(groupCourse: GroupCourseDueDateRow) {
+    const calendarEvent = this.buildCalendarEvent(groupCourse);
+
+    if (groupCourse.calendarEventId) {
+      await this.calendarRepository.updateCourseDueDateCalendarEvent(
```
Japrolol: This is not exactly an N+1 read, but it still does one DB write path per group. If this can run for larger bulk enrollments, can we batch the update/create work instead of awaiting each calendar event sync sequentially?

apps/api/src/courses/course.cron.ts
```diff
@@ -18,4 +18,11 @@ export class CourseCron {
       await this.courseService.sendOverdueCoursesEmails();
     });
   }
+
+  @Cron(CronExpression.EVERY_DAY_AT_9AM)
+  async notifyStudentsAboutCourseDueDatesPerDay() {
+    await this.tenantRunner.runForEachTenant(async () => {
+      await this.courseService.sendCourseDueDateReminders();
```
Japrolol, line 25: Do we need to run this tenant-by-tenant? If due reminders can be selected across tenants safely, one scheduler query would reduce cron overhead and avoid iterating every tenant just to discover whether it has work.

apps/api/src/courses/course.service.ts
```diff
@@ -2295,7 +2328,16 @@ export class CourseService {
     const groupIds = groupsToEnroll.map((group) => group.id);
     const groupInfoById = new Map(groupsToEnroll.map((group) => [group.id, group]));
 
-    const [course] = await this.db.select().from(courses).where(eq(courses.id, courseId));
+    const [course] = await this.db
+      .select({
+        authorId: courses.authorId,
+        title: sql<LocalizedText>`${courses.title}`,
```
Japrolol, line 2312: Same question as above. If the calendar event sync needs all localized course titles, the raw `LocalizedText` projection is probably correct. Can we make that intention explicit so this does not look like a missed localized-field lookup?

apps/api/src/courses/course.service.ts
```diff
@@ -4172,6 +4274,171 @@ export class CourseService {
...
+        html,
+      },
+      { tenantId: recipient.tenantId },
+    );
+  }
+
+  private async createCourseDueDateReminderAnnouncement(recipient: CourseDueDateReminderRecipient) {
+    const language = recipient.defaultEmailSettings.language;
+
+    const { heading, paragraphs } = getCourseDueDateReminderEmailTranslations(
```
Japrolol: Could this fan-out go through our CQRS/event-listener flow instead of sending emails directly from the cron path? Even with batching, SMTP latency or retries keep the tenant cron open. Publishing an event and handling the email side effect in a listener would keep the scheduler lighter and align with the rest of the app.

apps/api/src/courses/group-course-due-date-calendar.service.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    const calendarEventFields = this.buildCalendarEventFields(input);
+    const [existingCalendarEvent] = await dbInstance
+      .select({ id: calendarEvents.id })
+      .from(calendarEvents)
+      .where(and(eq(calendarEvents.uid, calendarEventFields.uid), isNull(calendarEvents.deletedAt)))
+      .orderBy(desc(calendarEvents.createdAt))
+      .limit(1);
+
+    const [syncedCalendarEvent] = existingCalendarEvent
```
Japrolol: Can we avoid the select-before-write here? Since `uid` identifies the course/group due-date event, an `insert ... on conflict do update` path would reduce one query per synced group and make the operation more concurrency-safe.

apps/api/src/courses/course.service.ts
its-gabo: I get the concern, but I think the current typing already makes that intention explicit. This field is typed as `LocalizedText`, specifically because the calendar event sync needs the full translated title map rather than one localized display value.

We also use the same `LocalizedText` type in plenty of places to represent full translation maps, so this feels consistent with the existing codebase convention.

Example: storage/schema/index.ts

apps/api/src/courses/group-course-due-date-calendar.service.ts
its-gabo: I don’t think `on conflict` fits this model as-is. `uid` is not unique today, and that is intentional because cancelled due-date events are soft-deleted and kept in `calendar_events`.

The current lookup only updates an active event: `uid = ... AND deleted_at IS NULL`

Japrolol: UID should be unique. If we want to re-use same UID let's update existing record and increment sequence

Edit: Additionally, if we add a unique constraint, then we can add `deletedAt: null` on insert and bump the sequence here to stick with iCalendar semantics, which 
1. Reduces query count
2. Keeps iCalendar semantics
3. Works the same way as it does now

apps/api/src/calendar/calendar.repository.ts
its-gabo: Same thing as below

apps/api/src/courses/course.cron.ts
its-gabo: Yes, this could be a valid concern if our tenant RLS/cross-tenant query approach was just better. For now I’d stick to our convention of running this per tenant, since it keeps the flow safely scoped by tenant context, including cases like global settings resolution.

This is also a widely used convention in our app for tenant-scoped background work.

Japrolol: I understand, I wrote this because I tried an approach, where I fetched only the tenants which had scheduled announcements and only ran `runForTenant` with them and it worked very well. But for now we can keep this approach if you want, although as I said in other comment, i'd publish this as an event using our `outboxPublisher` .

apps/api/src/storage/schema/index.ts
```diff
@@ -443,7 +443,7 @@ export const calendarEvents = pgTable(
       table.startsAt,
       table.endsAt,
     ),
-    uidIdx: index("calendar_events_tenant_uid_idx").on(table.tenantId, table.uid),
+    uidUniqueIdx: uniqueIndex("calendar_events_uid_unique_idx").on(table.uid),
```
Japrolol: Don't we want a uniqueIdx on uid and tenant, instead of just uid?

apps/api/src/calendar/calendar.repository.ts
```diff
@@ -42,6 +43,12 @@ import type {
   SupportedLanguages,
 } from "@repo/shared";
 
+type CourseDueDateCalendarEventUpsertInput = {
```
Japrolol: Can we put this in separate type file?

apps/api/src/common/helpers/sqlHelpers.ts
```diff
@@ -68,6 +68,16 @@ export function buildJsonbFieldWithMultipleEntries(entries: Partial<Record<strin
   return sql`jsonb_build_object(${sql.join(pairs, sql`, `)})`;
 }
 
+export function mergeJsonbField(existingField: any, incomingField: any) {
```
Japrolol: Not sure any is the right choice here, can we create a type like:
`type SqlExpression<T = unknown> = SQL<T> | AnyPgColumn;`

And use it here?

apps/api/src/courses/course.cron.ts
```diff
@@ -19,7 +19,7 @@ export class CourseCron {
     });
   }
 
-  @Cron(CronExpression.EVERY_DAY_AT_9AM)
+  @Cron(CronExpression.EVERY_5_SECONDS)
```
Japrolol: I believe you forgot to change back the value from testing

Issue #1121, opened 2025-12-19
Fix the last item on the list of lessons in the course details


marektro:
Each lesson has a number and a line leading to the next lesson. This vertical line should not exist on the last item component. It suggests there is another lesson, but this one is the last one and it should be visually clear for the user.

<img width="1170" height="698" alt="Image" src="https://github.com/user-attachments/assets/b783d791-8e5e-459c-8cfd-d683362dc3bf" />

its-gabo:
It's been fixed along the way

PR #1592, opened 2026-06-05
fix: hide the connector line on the last chapter step indicator

exAdmos:
## Issue
[#1121](https://github.com/Selleo/mentingo/issues/1121)

## Overview 
Added a `showConnector` prop to `ChapterCounter` (defaults to `true`) and an `isLast` prop to `CourseChapter` (defaults to `false`). When mapping chapters in `ChapterListOverview`, the last item receives `isLast={true}`, which passes `showConnector={false}` down to `ChapterCounter` and suppresses the `::after` vertical line.

## Business Value
Users browsing a course outline get a clear visual signal that they have reached the end of the chapter list. Without this fix the trailing line creates ambiguity.

## Screenshot
image


Japrolol (approved):

typeWolffo (approved):
:shipit:

PR #1596, opened 2026-06-08
fix: correct signup redirection for imported users

Japrolol:
## Issue(s)
[1594](https://github.com/Selleo/mentingo/issues/1594)

 ## Overview
  

  Fixes post-auth redirect handling after account creation and other authentication success flows.

  Changes include:

  - Makes handleAuthSuccess async so callers can wait for current-user query invalidation.
  - Updates login, registration, create-password, and magic-link mutations to await auth success handling.
  - Keeps auth state updates centralized in handleAuthSuccess.
  - Removes duplicated magic-link auth state/query handling and reuses the shared auth success helper.
  - Relies on current-user query invalidation instead of manually setting current-user query data, allowing the existing MFA guard and navigation-history flow to decide the
    final redirect.

  ## Business Value
  

  Imported users who create their password are redirected reliably without getting stuck on the login page until refresh.

  This preserves the existing MFA and navigation-history routing behavior while making authentication transitions more consistent across login, registration, magic-link, and
  account creation flows.

### DIFF
--- a/apps/web/app/api/mutations/helpers/handleAuthSuccess.ts
+++ b/apps/web/app/api/mutations/helpers/handleAuthSuccess.ts
@@ -24,7 +24,7 @@ type HandleAuthSuccessOptions = {
   setHasVerifiedMFA: (value: boolean) => void;
 };
 
-export function handleAuthSuccess({
+export async function handleAuthSuccess({
   user,
   setLoggedIn,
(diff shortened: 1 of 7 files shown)


its-gabo (approved):

its-gabo:
Let's just fix the not working E2E

Issue #1502, opened 2026-05-13
Add skeleton loading for course list


maciejhubert:
When course images are loading, it does not look appealing:

<img width="1654" height="987" alt="Image" src="https://github.com/user-attachments/assets/c33872c8-132f-4956-a73d-0419fc67eacf" />

Let's add skeleton loading until all the images are loaded.

its-gabo:
I'll also improve the loading speed of the courses page with infinite query

PR #1612, opened 2026-06-11
chore: improve modern course layout performance with faster loading

its-gabo:
## Issue(s)
- #1502 

## Overview

Improves the modern course layout loading strategy by replacing eager course/category prefetching with paginated, on-demand data loading.

- Adds a public API endpoint for available course categories, filtered by the same course filters used by available courses.
- Adds infinite query hooks for available courses, student courses, categories, and available course categories.
- Loads modern course rows by category in smaller pages and prefetches the next course/category pages as users approach them.
- Replaces the full-page loader with hero and course-row skeleton states.
- Memoizes modern course carousels/cards and exposes carousel viewport refs needed for intersection-based prefetching.
- Centralizes course list cache invalidation after course create/update flows.
- Updates generated Swagger and web API client artifacts for the new course category endpoint.

## Business Value

The modern courses page becomes faster to open and less expensive to render because it no longer fetches large course/category lists upfront. End users see meaningful skeleton states earlier, can start browsing sooner, and additional course rows load progressively as they scroll or navigate carousels.

This also keeps course list data fresher after course mutations by invalidating all related list queries from one shared utility.

## Screenshots / Video
https://github.com/user-attachments/assets/ecd080ca-1320-4e5f-afd6-df0508a281f9

### DIFF
--- a/apps/api/src/courses/course.service.ts
+++ b/apps/api/src/courses/course.service.ts
@@ -328,7 +328,7 @@ export class CourseService {
           );
           return { ...item, thumbnailUrl: signedUrl, authorAvatarUrl: authorAvatarSignedUrl };
         } catch (error) {
-          console.error(`Failed to get signed URL for ${item.thumbnailUrl}:`, error);
+          this.logger.error(`Failed to get signed URL for ${item.thumbnailUrl}:`, error);
           return item;
         }
       }),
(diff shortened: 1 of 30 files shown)


apps/api/src/courses/course.service.ts
```diff
@@ -646,6 +647,19 @@ export class CourseService {
...
 
+  private async getSignedCourseThumbnailUrl(
+    thumbnailReference: string | null,
+  ): Promise<string | null> {
+    if (!thumbnailReference) return thumbnailReference;
+
+    try {
+      return await this.fileService.getFileUrl(thumbnailReference);
+    } catch (error) {
+      console.error(`Failed to get signed URL for ${thumbnailReference}:`, error);
```
Japrolol: can we use logger instead of console.error?

Japrolol (approved):

PR #1620, opened 2026-06-11
chore: improve AI-powered course generation quality and reliability

Japrolol:
## Issue(s)
[1587](https://github.com/Selleo/mentingo/issues/1587)

## Overview
  

  Adds background sync for generated Luma courses into the Core curriculum.

  - Adds persisted Luma course generation sync state, queue processing, migrations, RLS, and websocket status updates.
  - Imports generated chapters, lessons, quizzes, AI mentor lessons, and ready lesson assets into Core curriculum structures.
  - Adds frontend sync/retry/dismiss handling for generated course imports.
  - Keeps curriculum editing locked while generation/sync is active and clears the lock on success, failure, or dismissal.
  - Adds typed shared constants for course generation sync statuses, stream event types, and message keys.
  - Tightens frontend stream parsing to use typed course generation events instead of broad fallbacks.

  ## Business Value
  

  Users can generate a course with Luma and have the result reliably materialized into the editable Mentingo curriculum. Background sync keeps the UI responsive, exposes
  recovery actions for failed imports, and prevents partial/generated content from leaving the curriculum in a blocked state.

### DIFF
--- a/apps/api/src/luma/luma-course-generation-sync.service.ts
+++ b/apps/api/src/luma/luma-course-generation-sync.service.ts
@@ -43,12 +43,12 @@ export class LumaCourseGenerationSyncService {
     try {
       await this.lumaCourseGenerationSyncQueueService.enqueueSyncJob({ courseId, currentUser });
     } catch (error) {
-      const lastError = this.sanitizeError(error);
+      const lastError = this.getErrorMessageForStorage(error);
       const failed = await this.lumaCourseGenerationSyncRepository.markFailed(courseId, lastError);
       const serialized = this.serialize(failed);
       this.publishStatusChanged(courseId, serialized, currentUser.userId, {
         messageKey: LUMA_COURSE_GENERATION_SYNC_MESSAGE_KEYS.FAILED,
-        error: lastError,
+        error: LUMA_COURSE_GENERATION_SYNC_MESSAGE_KEYS.FAILED,
       });
 
       return serialized;
@@ -82,7 +82,7 @@ export class LumaCourseGenerationSyncService {
 
       return serialized;
     } catch (error) {
-      const lastError = this.sanitizeError(error);
+      const lastError = this.getErrorMessageForStorage(error);
       this.logger.error(
         `Luma course generation sync failed for course ${data.courseId}: ${lastError}`,
         error instanceof Error ? error.stack : undefined,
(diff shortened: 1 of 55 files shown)


apps/api/src/luma/luma-course-generation-sync.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    if (!claimed) {
+      return this.serialize(
+        await this.lumaCourseGenerationSyncRepository.ensureNotStarted(courseId, null),
+      );
+    }
+
+    try {
+      await this.lumaCourseGenerationSyncQueueService.enqueueSyncJob({ courseId, currentUser });
+    } catch (error) {
+      const lastError = this.sanitizeError(error);
```
its-gabo: Is there a need to save last error. It's always the same error no matter what as sanitizeError returns `LUMA_COURSE_GENERATION_SYNC_MESSAGE_KEYS.FAILED`

apps/api/src/luma/luma-generated-course-import.service.ts
```diff
@@ -0,0 +1,760 @@
...
+    const contextIngestions: LumaAiMentorContextIngestion[] = [];
+    const stats: LumaGeneratedCourseImportStats = { skippedAssetCount: 0 };
+
+    await this.db.transaction(async (trx) => {
+      const language = await this.getCourseBaseLanguage(courseId, trx);
+      await this.assertCourseHasNoChapters(courseId, trx);
+      const assetMap = new Map(bundle.assets.map((asset) => [asset.assetId, asset]));
+
+      for (const [chapterDisplayIndex, chapter] of this.sortChapters(bundle.course).entries()) {
+        const chapterId = await this.insertChapter({
```
its-gabo, line 89: I'm wondering if using methods from `adminLessonService` wouldn't be a good idea here, as we will need to only care about changing it in one place. 

It's up to you tho

its-gabo (approved):
Everythings good, just 2 questions

apps/api/src/luma/luma-course-generation-sync.service.ts
Japrolol: Good catch, I'll keep the field and store errors there and in order not to reveal too much information to FE, I'll keep the `LUMA_COURSE_GENERATION_SYNC_MESSAGE_KEYS.FAILED" for it

apps/api/src/luma/luma-generated-course-import.service.ts
Japrolol: I will update for chapters, but I cannot do for lessons, because our current service is more user facing.

its-gabo (approved):

PR #1646, opened 2026-06-18
chore: improve user import reliability and performance

its-gabo:
## Overview

Improves the admin users import flow by moving import-specific logic into a dedicated `UserImportService` and repository, with bulk creation for users, roles, settings, onboarding, create-password tokens, user details, and group assignments.

The import now validates each row before creation and returns both imported and skipped rows. Skipped rows include explicit reasons for existing users, duplicate emails in the file, invalid roles, unavailable Trainer role setup, and unknown groups. If no rows are valid, the API returns a structured import failure response instead of partially proceeding.

The import side effects now go through outbox events:

- imported users receive invite emails through a batched notification handler
- a users import activity log is recorded with imported/skipped counts and row context

The web import result modal was also refreshed to show processed row count, imported/skipped tabs with counts, scrollable tables, localized skip reasons, and clearer empty states. API schemas and the generated web client were updated for the changed response shape.

## Business Value

Admins get clearer feedback after bulk user imports, including which users were created and why specific rows were skipped. This reduces manual troubleshooting and makes large imports safer to run.

The backend changes make the import process more reliable and auditable by batching database work, limiting invite email concurrency, and recording a dedicated activity log entry for each import. The result is a smoother onboarding workflow for organizations importing many users at once.

## Screenshots / Video

https://github.com/user-attachments/assets/03183a14-9e8c-450f-8e42-a7288aa9d5d6


apps/api/src/group/group.service.ts
```diff
@@ -49,6 +49,11 @@ import type {
 } from "src/group/group.types";
 import type { UserResponse } from "src/user/schemas/user.schema";
 
+type UserGroupAssignmentResult = {
```
Japrolol: Please extract to separate file

apps/api/src/group/group.service.ts
```diff
@@ -418,90 +423,139 @@ export class GroupService {
...
-          })
-          .where(
-            and(
-              eq(studentCourses.studentId, userId),
-              inArray(studentCourses.enrolledByGroupId, removedGroupIds),
+      await this.syncUserGroupEnrollment(userId, groupIdsToAssign, groupIdsToRemove, trx);
+
+      if (actor && groupIdsToAssign.length) {
+        await Promise.all(
+          groupIdsToAssign.map((groupId) =>
```
Japrolol: Would it be a good idea to use our `processInBatches` helper?

Japrolol (approved):

Japrolol (approved):

PR #1665, opened 2026-06-22
chore: allow users to have separate accounts with the same email across tenants

Japrolol:
## Issue(s)
[1655](https://github.com/Selleo/mentingo/issues/1655)

 ## Overview
  

  Allows users to have accounts with the same email address on different tenants.

  Changes include:

  - Replaces the global users.email unique constraint with a tenant-scoped unique index on (tenant_id, email).
  - Updates registration, admin user creation, and user import duplicate-email checks to use tenant-scoped DB context instead of DB_ADMIN.
  - Ensures user import validation runs duplicate-email lookup inside the same scoped transaction as the rest of import validation.
  - Adjusts the tenant RLS interceptor to preserve request execution inside runWithTenant.
  - Adds focused e2e coverage for same-email users across tenants, tenant-specific auth/token behavior, admin creation, and import duplicate lookup.

  ## Business Value
  

  Users can now belong to multiple tenant environments using the same email address, while each tenant still prevents duplicate accounts internally. This supports multi-tenant organizations more naturally
  and avoids blocking legitimate users who need access to separate tenant spaces with the same identity.

### DIFF
--- a/apps/api/src/storage/migrations/0152_tenant_scoped_user_email_unique.sql
+++ b/apps/api/src/storage/migrations/0152_tenant_scoped_user_email_unique.sql
@@ -1,0 +1,3 @@
+ALTER TABLE "users" DROP CONSTRAINT "users_email_unique";--> statement-breakpoint
+ALTER TABLE "courses" ALTER COLUMN "settings" SET DEFAULT '{"lessonSequenceEnabled":false,"quizFeedbackEnabled":true,"certificateSignature":null,"certificateFontColor":null,"certificateValidity":null,"videoCompletionTrackingEnabled":true}'::jsonb;--> statement-breakpoint
+CREATE UNIQUE INDEX IF NOT EXISTS "users_tenant_id_email_unique_idx" ON "users" USING btree ("tenant_id","email");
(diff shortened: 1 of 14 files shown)


its-gabo (approved):

apps/api/src/storage/migrations/0152_tenant_scoped_user_email_unique.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "users" DROP CONSTRAINT "users_email_unique";--> statement-breakpoint
+ALTER TABLE "courses" ALTER COLUMN "settings" SET DEFAULT '{"lessonSequenceEnabled":false,"quizFeedbackEnabled":true,"certificateSignature":null,"certificateFontColor":null,"certificateValidity":null,"videoCompletionTrackingEnabled":true}'::jsonb;--> statement-breakpoint
```
its-gabo, line 2: Is it supposed to be here?

Japrolol: Yes, as we set bew default value previously for settings and the schema updated but migration wasn't generated.

Issue #1678, opened 2026-06-25
Fix correct sentence missing in gap filling


maciejhubert:
> I see that this is not fixed, I am creating a separate issue for it:
> 
> Also, correct sentence is missing in gap filling:
> 
> <img width="1914" height="1030" alt="Image" src="https://github.com/user-attachments/assets/92655d31-561a-4f2a-8485-f03c10311a19" />
> 
>  

 _Originally posted by @maciejhubert in [#1639](https://github.com/Selleo/mentingo/issues/1639#issuecomment-4796900983)_

maciejhubert:
I see 2 issues.

Issue 1 - the correct answer is still missing for the text gap fill:

<img width="1034" height="897" alt="Image" src="https://github.com/user-attachments/assets/e0dd64d7-4c82-4278-8106-f7d9edbe5981" />

Issue 2 - in [this case](https://learn.selleo.com/course/8c6h8-warsztaty-modele-prawdopodobienstwa-w-grach-i-decyzjach/lesson/a2b2c4c8-8a21-42c9-85fc-e87c012f2673) I have disabled quiz feedback, but I can still see the feedback after retaking it. Not sure if this is regression, but let's fix it.

PR #1682, opened 2026-06-25
fix: display the correct sentence in fill-in-the-blank exercises

its-gabo:
## Issue(s)
- #1678 

## Overview
Fixes missing correct-sentence feedback for failed fill-in-the-blanks quiz questions.

The learner result view now uses the stored solution explanation when available and falls back to deriving the correct sentence from the question description plus correct answer options. This applies to both text-entry and drag-and-drop fill-in-the-blanks questions.

This also updates seed gap-filling descriptions to use the current blank-answer marker format and hardens global search indexing so lesson save does not fail when a localized JSONB field is not an object.

## Business Value
Learners who answer a gap-filling question incorrectly can see the intended completed sentence after submission, making feedback actionable instead of only showing selected or typed words.

For L&D teams, this improves quiz review quality and keeps existing or imported quiz content usable even when stored solution text is missing.

## Screenshots / Video

https://github.com/user-attachments/assets/11f722a0-d70f-4285-a235-29d5684e8de9

### DIFF
--- a/apps/web/app/modules/Courses/Lesson/Question/FillInTheBlanks/correctSentence.ts
+++ b/apps/web/app/modules/Courses/Lesson/Question/FillInTheBlanks/correctSentence.ts
@@ -1,17 +1,11 @@
+import { escape } from "lodash-es";
+
 import {
   normalizeBlankAnswerLineBreaks,
   splitByBlankAnswerMarkers,
 } from "~/utils/blankAnswerMarkers";
 
 import type { QuizQuestion, QuizQuestionOption } from "../types";
-
-const escapeHtml = (value: string) =>
-  value
-    .replace(/&/g, "&amp;")
-    .replace(/</g, "&lt;")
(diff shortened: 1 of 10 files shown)


apps/web/app/modules/Courses/Lesson/Question/FillInTheBlanks/correctSentence.ts
```diff
@@ -0,0 +1,62 @@
+import {
+  normalizeBlankAnswerLineBreaks,
+  splitByBlankAnswerMarkers,
+} from "~/utils/blankAnswerMarkers";
+
+import type { QuizQuestion, QuizQuestionOption } from "../types";
+
+const escapeHtml = (value: string) =>
```
Japrolol: can we use `escape()` from `lodash`?

Japrolol (approved):

PR #1686, opened 2026-06-27
feat: allow admins to duplicate course

its-gabo:
## Issue(s)
- #1685 

## Overview
Adds course duplication for admins. The flow creates a draft placeholder immediately, queues the full course copy in BullMQ, and updates the UI through websocket notifications plus job-status polling so refreshes and missed socket events are handled.

The duplication copies course content, chapters, supported lessons, questions, answers, resources, AI mentor data, and SCORM packages while avoiding copied primary keys and source-owned identifiers. Duplicated course titles receive a localized copy suffix based on the course/title language.

## Business Value
Admins can quickly reuse existing courses as editable drafts instead of rebuilding course structure and content manually. This reduces setup time for repeated training programs, localized variants, and course iterations while keeping the original course unchanged.

## Screenshots / Video

https://github.com/user-attachments/assets/de60e74a-54e3-4c65-83b1-9efadffef055


apps/api/src/courses/course-duplication.service.ts
```diff
@@ -0,0 +1,199 @@
...
+import { SearchIndexService } from "src/global-search/search-index.service";
+import { DB } from "src/storage/db/db.providers";
+import { courses, coursesSummaryStats } from "src/storage/schema";
+import { WsGateway } from "src/websocket";
+
+import type { CurrentUserType } from "src/common/types/current-user.type";
+import type { DuplicateCourseResponse } from "src/courses/schemas/courseDuplication.schema";
+import type { CourseDuplicationJobData } from "src/queue";
+
+type LocalizedStringMap = Record<string, string>;
```
Japrolol: Can we take this out from `service` to separate file?

apps/api/src/courses/course-duplication.service.ts
```diff
@@ -0,0 +1,199 @@
...
+
+const COURSE_DUPLICATION_COPY_SUFFIX: Record<SupportedLanguages, string> = {
```
Japrolol: Same as above

apps/api/src/courses/course-duplication.service.ts
Japrolol, line 1: Can we move sql to repository instead of keeping it in service?

apps/api/src/courses/course-duplication.service.ts
```diff
@@ -0,0 +1,199 @@
...
+      }),
+    );
+  }
+
+  private getCopySuffix(language: SupportedLanguages) {
+    return COURSE_DUPLICATION_COPY_SUFFIX[language];
+  }
+
+  private toLocalizedStringMap(value: unknown): LocalizedStringMap {
+    if (!value || typeof value !== "object" || Array.isArray(value)) return {};
```
Japrolol: Maybe we can add `.$type<LocalizedText>()` to courses title etc in db schema to avoid these weird type normalizations

Japrolol (approved):

Issue #1641, opened 2026-06-17
Improve video player experience


maciejhubert:
Let's improve a few things in the video player. If some are more complex than other, feel free to separate into multiple tasks.

Implement shortcuts. YT is a good benchmark:
* F for full screen
* left/right arrow for 5 second rewind/forward
* space for pause/play
* up/down arrow for volume change
* M for mute
* 0 for setting the video to 0:00
* let's give that thought 10 minutes (I know we spend too much already on quirks like autoplay) - can we focus on player after opening a lesson to make it smoother? If not, let's ignore it

Japrolol:
Let's also fix here the grid layout for 2x2 in live training

PR #1687, opened 2026-06-29
feat: add keyboard shortcuts for easier video playback control

its-gabo:
## Issue(s)
- #1641 

## Overview
Adds focused keyboard shortcuts to the shared lesson video player:

- `Space` / `Enter` toggles play and pause.
- Left and right arrows seek backward or forward by 5 seconds.
- Up and down arrows adjust volume.
- `M` toggles mute.
- `0` restarts the video from the beginning.
- `F` toggles fullscreen.

The video player now focuses itself after thumbnail activation so learners can use shortcuts immediately, while shortcut handling stays scoped to the player and avoids inputs, menus, and native player controls.

This also improves player control visibility in fullscreen and while interacting with the control bar, and updates the live training participant grid so four visible participant tiles render as a stable 2x2 layout.

## Business Value
Learners get a smoother video lesson experience with familiar playback shortcuts and more reliable fullscreen controls. This reduces friction during self-paced training, especially for longer video lessons where seeking, muting, and fullscreen playback are common actions.

Live training rooms are easier to scan in small group sessions because the four-participant layout is balanced instead of wrapping unpredictably.

### DIFF
--- a/apps/web/app/components/VideoPlayer/VideoPlayer.tsx
+++ b/apps/web/app/components/VideoPlayer/VideoPlayer.tsx
@@ -32,12 +32,14 @@ import { useVideoCoverageTracker } from "./useVideoCoverageTracker";
 
 import type { VideoAspectRatio } from "./aspectRatio";
 import type { VideoCoverageTrackingOptions } from "./videoCoverage.types";
+import type { KeyboardEvent } from "react";
 
 interface VideoPlayerProps {
   url: string;
   onAspectRatioChange?: (aspectRatio: VideoAspectRatio) => void;
   onEnded?: () => void;
   autoPlay?: boolean;
+  focusOnMount?: boolean;
   fill?: boolean;
   className?: string;
   provider: VideoProvider;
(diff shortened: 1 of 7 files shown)


apps/web/app/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -48,10 +50,76 @@ type VideoJSType = ReturnType<typeof videojs>;
...
+  ".mentingo-vjs-quality-selector",
+].join(",");
+const PLAYER_CONTROLS_TARGET_SELECTOR = [
+  ".vjs-control-bar",
+  ".vjs-control",
+  ".vjs-menu",
+  ".mentingo-vjs-quality-selector",
+].join(",");
+
+const clamp = (value: number, min: number, max: number) => Math.min(Math.max(value, min), max);
```
Japrolol: we can use `clamp()` from `lodash`

apps/web/app/components/VideoPlayer/VideoPlayer.tsx
```diff
@@ -188,10 +260,155 @@ export const VideoPlayer = ({
...
+        showControls();
+      }
+
+      if (isKeyboardShortcutIgnoredTarget(event.target, currentTarget)) return;
+
+      const currentPlayer = playerRef.current;
+      if (!currentPlayer || currentPlayer.isDisposed()) return;
+
+      if (
+        event.key !== " " &&
```
Japrolol: Would it be possible to put these event keys in one array and then just use `.includes()`. We could normalize all to lowercase in the stored array and then also transform event key to lowercase.

Japrolol (approved):

PR #1711, opened 2026-07-01
fix: video playback access for unregistered users

its-gabo:
## Overview

Fixes public video playback for unregistered users by authorizing lesson resources by the requested resource ID instead of relying on a single resource-entity row. This allows reused video resources to stream when any linked lesson is a public freemium content lesson and visitor course access is enabled.

The change also updates internal thumbnail access checks to evaluate all entities linked to a resource and allow the thumbnail when at least one linked entity grants access.

## Business Value

Unregistered learners can reliably watch public course videos in freemium lessons, including videos reused across multiple lessons. This keeps publicly accessible course content playable without weakening access checks for private lessons, quizzes, or disabled visitor access.

### DIFF
--- a/apps/api/src/file/thumbnail.service.ts
+++ b/apps/api/src/file/thumbnail.service.ts
@@ -75,18 +75,13 @@ export class ThumbnailService {
     if (!resourceId) throw new BadRequestException("Invalid sourceUrl");
 
     const [resource] = await this.db
-      .select({
-        ...getTableColumns(resources),
-        entityType: resourceEntity.entityType,
-        entityId: resourceEntity.entityId,
-      })
+      .select({ ...getTableColumns(resources) })
       .from(resources)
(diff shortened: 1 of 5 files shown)


apps/api/src/file/thumbnail.service.ts
```diff
@@ -152,6 +147,34 @@ export class ThumbnailService {
     }
   }
 
+  private async validateThumbnailAccessForResource(
+    resourceId: UUIDType,
+    currentUser: CurrentUserType | null,
+  ) {
+    const resourceLinks = await this.db
```
Japrolol, line 154: I am aware that we had DB queries inside this service previously, but since we already change stuff here, could we maybe extract them to `thumbnail.repository.ts`. Also I noticed we have missing translations in exceptions

Issue #1420, opened 2026-04-09
Improve design for changing course list layout


maciejhubert:
In admin's Settings -> Platform customization, let's make it more clear how each layout looks:

[Design layout toggle settings.zip](https://github.com/user-attachments/files/26591475/Design.layout.toggle.settings.zip)

<img width="1135" height="890" alt="Image" src="https://github.com/user-attachments/assets/2bea7e93-0dd5-4afd-a8df-a5c243a1c801" />

PR #1735, opened 2026-07-06
chore: improve design for changing course list layout (#1420)

exAdmos:
## Issue
[1420](https://github.com/Selleo/mentingo/issues/1420/)

## Overview 
Adds two cards with possible layouts and needed translations.

## Business Value
These changes make the UI more descriptive.

## Screenshot
image


apps/web/app/modules/Dashboard/Settings/components/admin/CourseListLayoutPreview.tsx
typeWolffo, line 1: Consider adding translations for all strings

typeWolffo: also @its-gabo @Japrolol do we have plans to add new colors to the config? I see that some of these are already used in the certificates

typeWolffo (changes requested):

apps/web/app/modules/Dashboard/Settings/components/admin/CourseListLayoutPreview.tsx
```diff
@@ -0,0 +1,106 @@
+interface CourseListLayoutPreviewProps {
+  type: "classic" | "modern";
+}
+
+export function CourseListLayoutPreview({ type }: CourseListLayoutPreviewProps) {
+  if (type === "modern") {
+    return (
+      <div className="bg-white rounded border border-[#e5e7eb] p-4 h-[400px] overflow-hidden">
+        {/* Hero Section Skeleton */}
```
Japrolol: Let's delete those `claude` comments, comments should only be used when they bring value. Here it's not meaningful.

apps/web/app/modules/Dashboard/Settings/components/admin/CourseListLayoutPreview.tsx
```diff
@@ -0,0 +1,106 @@
+interface CourseListLayoutPreviewProps {
+  type: "classic" | "modern";
```
Japrolol: We can create a separate file to have `classic` and `modern` in an enum in this format `const x = {...} as const`. Then using that enum let's get the type from it, which can then be used here.

apps/web/app/modules/Dashboard/Settings/components/admin/CourseListLayoutPreview.tsx
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export function CourseListLayoutPreview({ type }: CourseListLayoutPreviewProps) {
+  if (type === "modern") {
```
Japrolol: Use the previously stated enum

apps/web/app/modules/Dashboard/Settings/components/admin/CourseListLayoutPreview.tsx
```diff
@@ -0,0 +1,106 @@
...
+    return (
+      <div className="bg-white rounded border border-[#e5e7eb] p-4 h-[400px] overflow-hidden">
+        {/* Hero Section Skeleton */}
+        <div className="relative h-[140px] bg-gradient-to-br from-[#6b7280] to-[#9ca3af] rounded overflow-hidden mb-4">
```
Japrolol: Can we avoid hex values?

apps/web/app/modules/Dashboard/Settings/components/admin/CourseListLayoutPreview.tsx
```diff
@@ -0,0 +1,106 @@
...
+          <div className="absolute inset-0 bg-gradient-to-t from-black/40 via-black/20 to-transparent"></div>
+          <div className="absolute bottom-4 left-4 space-y-2">
+            <div className="h-[8px] w-[80px] bg-white/50 rounded"></div>
```
Japrolol: Let's avoid using pure pixel values, when a direct tailwind equivalent exists like `h-2`

apps/web/app/modules/Dashboard/Settings/components/admin/CoursesAccessibilityPreferences.tsx
```diff
@@ -42,14 +46,43 @@ export default function CoursesAccessibilityPreferences({
...
-      />
+      <div className="space-y-3">
+        <div className="space-y-0.5">
+          <p className="body-base-md">{t("adminPreferences.field.courseListLayout")}</p>
+          <p className="body-sm-md text-muted-foreground">
+            {t("adminPreferences.field.courseListLayoutDescription")}
+          </p>
+        </div>
+        <RadioGroup
+          value={globalSettings.modernCourseListEnabled ? "modern" : "classic"}
```
Japrolol: Let's extract this condition once at the top and re-use it.

apps/web/app/modules/Dashboard/Settings/components/admin/CoursesAccessibilityPreferences.tsx
```diff
@@ -42,14 +46,43 @@ export default function CoursesAccessibilityPreferences({
...
+            const current = globalSettings.modernCourseListEnabled ? "modern" : "classic";
+            if (val !== current) toggleModernCourseList();
+          }}
+          className="flex flex-col gap-4 sm:flex-row sm:gap-10 sm:justify-center"
+        >
+          {(["classic", "modern"] as const).map((layout) => (
+            <Label key={layout} htmlFor={layout} className="w-full cursor-pointer sm:flex-1">
+              <RadioGroupItem value={layout} id={layout} className="sr-only" />
+              <div
+                className={`border-2 rounded-lg p-3 transition-colors ${
```
Japrolol: Let's use `cn()` for conditional class names

apps/web/app/modules/Dashboard/Settings/components/admin/CoursesAccessibilityPreferences.tsx
```diff
@@ -42,14 +46,43 @@ export default function CoursesAccessibilityPreferences({
...
+                  (globalSettings.modernCourseListEnabled ? "modern" : "classic") === layout
+                    ? "border-primary bg-primary/5"
+                    : "border-neutral-200 hover:border-neutral-300"
+                }`}
+              >
+                <p className="text-base font-bold capitalize mb-1">{layout}</p>
+                <p className="text-sm text-muted-foreground mb-2">
+                  {layout === "classic"
```
Japrolol: use the previously specified enum

apps/web/app/modules/Dashboard/Settings/components/admin/CourseListLayoutPreview.tsx
Japrolol: Let's delete those `claude` comments, comments should only be used when they bring value. Here it's not meaningful.

apps/web/app/modules/Dashboard/Settings/components/admin/CoursesAccessibilityPreferences.tsx
```diff
@@ -42,14 +46,43 @@ export default function CoursesAccessibilityPreferences({
...
+        </div>
+        <RadioGroup
+          value={globalSettings.modernCourseListEnabled ? "modern" : "classic"}
+          onValueChange={(val) => {
+            const current = globalSettings.modernCourseListEnabled ? "modern" : "classic";
+            if (val !== current) toggleModernCourseList();
+          }}
+          className="flex flex-col gap-4 sm:flex-row sm:gap-10 sm:justify-center"
+        >
+          {(["classic", "modern"] as const).map((layout) => (
```
Japrolol: Can we use `Object.values` on the previously created enum?

apps/web/app/modules/Dashboard/Settings/components/admin/CourseListLayoutPreview.tsx
Japrolol: I would use the most similar tailwind values as this doesn't have to be 1:1 true to design color, this way we avoid extending the tailwind config.

Japrolol (changes requested):

maciejhubert:
@exAdmos 
Please consider adding a darker background for the selected option, just like in the following example:
image


Also let's add radio buttons to make sure it looks like options to pick from.

typeWolffo (approved):

Japrolol (approved):

PR #1738, opened 2026-07-07
fix: display correct translations for localized categories and groups

exAdmos:
## Issue(s)
[1578](https://github.com/Selleo/mentingo/issues/1578)

## Overview 
Adds missing labes

## Screenshots / Video
image

### DIFF
--- a/apps/api/src/seed/seed-helpers.ts
+++ b/apps/api/src/seed/seed-helpers.ts
@@ -5,13 +5,11 @@ import {
   SYSTEM_ROLE_SLUGS,
   SYSTEM_RULE_SET_SLUGS,
   SUPPORTED_LANGUAGES,
-  isSupportedLanguage,
   type SystemRoleSlug,
-  type LocalizedText,
 } from "@repo/shared";
 import { and, eq, sql } from "drizzle-orm/sql";
 
-import { buildJsonbField, buildJsonbFieldWithMultipleEntries } from "src/common/helpers/sqlHelpers";
+import { buildJsonbField } from "src/common/helpers/sqlHelpers";
 import { EnvRepository } from "src/env/repositories/env.repository";
 import { EnvService } from "src/env/services/env.service";
 import { LESSON_TYPES } from "src/lesson/lesson.type";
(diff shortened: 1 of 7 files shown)


apps/api/src/seed/seed-helpers.ts
```diff
@@ -34,6 +35,58 @@ import { StripeService } from "src/stripe/stripe.service";
 import type { DatabasePg, UUIDType } from "../common";
 import type { NiceCourseData } from "../utils/types/test-types";
 
+const CATEGORY_TRANSLATIONS: Record<string, Partial<Record<string, string>>> = {
```
typeWolffo: we already have type for that - try `Record<string, LocalizedText>` 😉

apps/api/src/seed/seed-helpers.ts
```diff
@@ -101,14 +154,19 @@ export async function createNiceCourses(
...
     const [category] = existingCategory
       ? [existingCategory]
       : await db
           .insert(categories)
           .values({
-            title: buildJsonbField(SUPPORTED_LANGUAGES.EN, courseData.category),
+            title: buildJsonbFieldWithMultipleEntries(categoryTranslations),
             baseLanguage: SUPPORTED_LANGUAGES.EN,
-            availableLocales: [SUPPORTED_LANGUAGES.EN],
+            availableLocales: Object.keys(categoryTranslations) as SupportedLanguages[],
```
typeWolffo: instead of type assertion, filter with the type guard `.filter(isSupportedLanguage)`

typeWolffo (changes requested):

typeWolffo (approved):
:shipit:

apps/api/src/swagger/api-schema.json
```diff
@@ -16544,6 +16544,10 @@
                       "const": "announcement.delete",
                       "type": "string"
                     },
+                    {
+                      "const": "email_template.manage",
+                      "type": "string"
+                    },
```
its-gabo: I believe it shouldn't be here

its-gabo (changes requested):

its-gabo:
Also are you sure that the task was about it? I believe it was about displaying properly translated entries from database.

its-gabo (approved):

PR #1753, opened 2026-07-09
feat: add in-app mention notifications and improve the notification system

pkrzysiekk:
## Issue(s)
[1617](https://github.com/Selleo/mentingo/issues/1617)

## Overview 
When discusions are enabled in the course section, mentioning user creates an event that notifies mentioned user in an in-app notiification center

Since the notification logic changed a bit (admin shouldn't recieve it's own in-app notifications) I changed one announcement controller test because it required admin to see their own notifications.

## Business Value
It helps mentioned users to stay in touch with users that mentioned them in various course discussions.


mountain-bit (commented):
Overall, the changes look good. My only concern is the localization approach. I think it could become difficult to maintain if we need to update translations or add support for another language in the future.

apps/api/src/announcements/announcements.repository.ts
```diff
@@ -143,6 +160,17 @@ export class AnnouncementsRepository {
         announcementId: announcement.id,
       });
     }
+    if (input.usersToNotify) {
+      await this.db
+        .insert(userAnnouncements)
+        .values(
+          input.usersToNotify!.map((userId) => ({
```
its-gabo: I believe that there's no need for the `!`, because we have and if that checks if it exists. I would also create a variable for for the output of this map and then just pass the created variable to values

```suggestion
          input.usersToNotify.map((userId) => ({
```

apps/api/src/course-chat/handlers/course-chat-mention-email.handler.ts
```diff
@@ -56,6 +65,27 @@ export class CourseChatMentionEmailHandler
...
+      const mentioningUser = await this.userService.getUserById(currentUser.userId);
+      const mentioningUserFullName = mentioningUser.firstName + " " + mentioningUser.lastName;
+      const courseData = await this.courseService.getCourseEmailData(courseId);
+
+      await this.announcementRepository.createAnnouncement({
+        groupId: null,
+        title: GetLocalizedUserMentionTitleAnnouncement(mentioningUserFullName),
+        content: GetLocalizedUserMentionContentAnnouncement(courseData.courseName),
+        baseLanguage: "en",
+        availableLocales: ["en", "pl", "de", "lt", "cs", "es"],
```
its-gabo:
```suggestion
        availableLocales: [...Object.values(SUPPORTED_LANGUAGES)],
```

apps/api/src/course-chat/handlers/course-chat-mention-email.handler.ts
```diff
@@ -56,6 +65,27 @@ export class CourseChatMentionEmailHandler
...
+        authorId: currentUser.userId,
+        status: ANNOUNCEMENT_STATUSES.PUBLISHED,
+        scheduledAt: null,
+        publishedAt: null,
+        sendEmail: false,
+        emailTemplate: "default",
+        sourceType: "manual",
```
its-gabo: Please use `ANNOUNCEMENT_SOURCE_TYPES` and add `COURSE_CHAT` to this enum

apps/api/src/course-chat/handlers/course-chat-mention-email.handler.ts
```diff
@@ -56,6 +65,27 @@ export class CourseChatMentionEmailHandler
...
+        title: GetLocalizedUserMentionTitleAnnouncement(mentioningUserFullName),
+        content: GetLocalizedUserMentionContentAnnouncement(courseData.courseName),
+        baseLanguage: "en",
+        availableLocales: ["en", "pl", "de", "lt", "cs", "es"],
+        authorId: currentUser.userId,
+        status: ANNOUNCEMENT_STATUSES.PUBLISHED,
+        scheduledAt: null,
+        publishedAt: null,
+        sendEmail: false,
+        emailTemplate: "default",
```
its-gabo: Here use `ANNOUNCEMENT_EMAIL_TEMPLATES`

apps/api/src/course-chat/chat-mention-localizations/chat-mention-content-localization.ts
```diff
@@ -0,0 +1,13 @@
+import type { SupportedLanguages } from "node_modules/@repo/shared/dist/index.cjs";
+
+export function GetLocalizedUserMentionContentAnnouncement(courseName: string) {
```
its-gabo: Please make the function name start with lower-cased letter

apps/api/src/course-chat/chat-mention-localizations/chat-mention-title-localization.ts
```diff
@@ -0,0 +1,13 @@
+import type { SupportedLanguages } from "@repo/shared/";
+
+export function GetLocalizedUserMentionTitleAnnouncement(mentioningUserFullName: string) {
```
its-gabo: Please make the function name start with lower-cased letter

apps/api/src/course-chat/chat-mention-localizations/chat-mention-content-localization.ts
```diff
@@ -0,0 +1,13 @@
+import type { SupportedLanguages } from "node_modules/@repo/shared/dist/index.cjs";
```
its-gabo:
```suggestion
import type { SupportedLanguages } from "@repo/shared";
```

apps/api/src/announcements/announcements.controller.ts
```diff
@@ -65,8 +65,10 @@ export class AnnouncementsController {
     @Query("status") status?: AnnouncementStatus,
     @Query("page") page?: number,
     @Query("perPage") perPage?: number,
+    @CurrentUser() currentUser?: CurrentUserType,
   ): Promise<PaginatedResponse<AllAnnouncements>> {
     const announcements = await this.announcementsService.getAllAnnouncements(
+      currentUser!,
```
its-gabo: We can make the current user not optional therefore we would be able to just pass it without `!` 

```suggestion
    @CurrentUser() currentUser: CurrentUserType,
): Promise<PaginatedResponse<AllAnnouncements>> {
const announcements = await this.announcementsService.getAllAnnouncements(
      currentUser,
```

its-gabo (changes requested):
Please avoid filtering announcements by author or inferring visibility from userAnnouncements, as broadcast/system announcements may also have user-announcement records for tracking if the announcement is read.

Instead, model the intended audience explicitly (e.g. audience: "all_users" | "selected_users" - it can replace `isEveryone` field).

Also please remove the redundant changes that were caused by formatting.

its-gabo (changes requested):
You'll need to make the migration in 3 steps (3 migrations):
1. add audience column
2. set audience to all_users where isEveryone is true (custom migration)
3. remove isEveryone column

You need to do this, because we will lose the state that we had. If I'm not wrong then your current migration will change true to "true" and false to "false", which is not what we want

packages/email-templates/tsup.config.bundled_op5l35bgbz7.mjs
```diff
@@ -12,44 +12,46 @@ var traverse = _traverse.default;
 var generate = _generate.default;
```
its-gabo: Let's remove this file

apps/web/app/api/generated-api.ts
```diff
@@ -1,14 +1,14 @@
...
-/*
- * ---------------------------------------------------------------
- * ## THIS FILE WAS GENERATED VIA SWAGGER-TYPESCRIPT-API        ##
- * ##                                                           ##
- * ## AUTHOR: acacode                                           ##
- * ## SOURCE: https://github.com/acacode/swagger-typescript-api ##
- * ---------------------------------------------------------------
- */
-
+/* eslint-disable */
```
its-gabo: Let's regenerate this file using `pnpm generate:client`

its-gabo (changes requested):
Looks fine, there are just many files that just got formatted even if you didn't change them - let's unstage them

These files:
- `packages/prompts/src/templates/security-and-rag-block-prompt.yaml`
- `packages/prompts/src/templates/summary-prompt.yaml`
- `packages/prompts/src/templates/welcome-prompt.yaml`
- `packages/scorm-export-runtime/scripts/copy-player-assets.mjs`
- `packages/ui/turbo/generators/templates/component.hbs`
- `docker-compose.langfuse.yml`

packages/email-templates/plugin.mjs
```diff
@@ -26,21 +26,20 @@ const generateIndexContent = () => {
         [
```
its-gabo, line 26: I'm sorry I missed this file 😅 

Can you unstage it too?

its-gabo (approved):

Japrolol (approved):

PR #1760, opened 2026-07-10
chore: improve activity logs for article language changes

pkrzysiekk:
## Issue(s)
[1152](https://github.com/Selleo/mentingo/issues/1152)

## Overview 
Added specific actions (Delete and Add) in activity logs when article language or article section language is created or deleted. Before those operations were logged as simply Update event which wasn't that informative. Log information were also adjusted, removed unnecessary before state when creating.
## Business Value 

Adding specific action on Delete and Create action makes it easier for admin to track changes of what happened to the articles.


mountain-bit (commented):
Looks good to me.

Pieselak (approved):

Japrolol (approved):

PR #1768, opened 2026-07-14
chore: add activity logging and test coverage for SCORM and development paths

mountain-bit:
## Issue(s)
[1477](https://github.com/Selleo/mentingo/issues/1477)

## Overview 

Extends the platform's Activity Logs to cover SCORM packages and Learning Paths. Every key operation — create, update, delete, enroll, start, and complete — is now automatically recorded and visible to administrators. Changes include dispatching missing events (CreateScormEvent, DeleteScormEvent), adding translations, and covering the new behavior with unit and E2E tests.

## Business Value

Administrators gain full audit trail for Learning Paths and SCORM — who enrolled students, who deleted a package, when a learner started or completed a path. 

## Screenshots / Video
image
image


pkrzysiekk (approved):
Everything looks fine

AdamStajer07 (approved):
everything fine

apps/api/src/activity-logs/handlers/learning-path-activity.handler.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  private async handleCompleteLearningPath(event: CompleteLearningPathEvent) {
+    await this.activityLogsService.recordActivity({
+      actor: event.completeData.actor,
+      tenantId: event.completeData.actor.tenantId,
+      operation: ACTIVITY_LOG_ACTION_TYPES.COMPLETE_LEARNING_PATH,
+      resourceType: ACTIVITY_LOG_RESOURCE_TYPES.LEARNING_PATH,
+      resourceId: event.completeData.learningPathId,
+      context: {
+        learningPathId: event.completeData.learningPathId,
```
its-gabo: I believe this not necessary here, as the `learningPathId` is passed to `resourceId`

apps/api/src/activity-logs/handlers/scorm-activity.handler.ts
```diff
@@ -0,0 +1,124 @@
...
+    await this.activityLogsService.recordActivity({
+      actor: event.playData.actor,
+      tenantId: event.playData.actor.tenantId,
+      operation: ACTIVITY_LOG_ACTION_TYPES.PLAY_SCORM,
+      resourceType: ACTIVITY_LOG_RESOURCE_TYPES.SCORM,
+      resourceId: event.playData.scormId,
+      context:
+        event.playData.userId !== event.playData.actor.userId
+          ? { playedByUserId: event.playData.userId }
+          : null,
```
its-gabo: Same as above - is this possible to happen?

apps/api/src/activity-logs/handlers/learning-path-activity.handler.ts
```diff
@@ -0,0 +1,148 @@
...
+        ...(event.completeData.userId !== event.completeData.actor.userId
+          ? { completedByUserId: event.completeData.userId }
+          : {}),
```
its-gabo: Is it possible that such case happens?

apps/api/src/activity-logs/handlers/learning-path-activity.handler.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  private async handleStartLearningPath(event: StartLearningPathEvent) {
+    await this.activityLogsService.recordActivity({
+      actor: event.startData.actor,
+      tenantId: event.startData.actor.tenantId,
+      operation: ACTIVITY_LOG_ACTION_TYPES.START_LEARNING_PATH,
+      resourceType: ACTIVITY_LOG_RESOURCE_TYPES.LEARNING_PATH,
+      resourceId: event.startData.learningPathId,
+      context: {
+        learningPathId: event.startData.learningPathId,
```
its-gabo: I believe this not necessary here, as the learningPathId is passed to resourceId

apps/api/src/courses/course.service.ts
```diff
@@ -3254,9 +3266,19 @@ export class CourseService {
...
+
+    if (scormPackageToDelete) {
+      await this.outboxPublisher.publish(
+        new DeleteScormEvent({
+          scormId: scormPackageToDelete.id,
+          actor: currentUser,
+        }),
+        this.db,
+      );
+    }
```
its-gabo: I'd put it in transaction if possible, so we ensure that if the transaction fails we don't save the delete event

apps/api/src/learning-path/services/learning-path-course-sync.service.ts
```diff
@@ -404,16 +469,72 @@ export class LearningPathCourseSyncService {
...
+    if (!isStarted && !isCompleted) return;
+
+    let actor: ActorUserType | null = null;
+
+    try {
+      actor = await this.learningPathRepository.getStudentActorInfo(studentId);
+    } catch (error) {
+      console.error(`Failed to fetch actor info for student ${studentId}:`, error);
+      return;
+    }
```
its-gabo: I'd avoid this try/catch here

apps/api/src/learning-path/services/learning-path-course-sync.service.ts
```diff
@@ -404,16 +469,72 @@ export class LearningPathCourseSyncService {
...
+
+    if (isCompleted) {
+      await this.outboxPublisher.publish(
+        new CompleteLearningPathEvent({
+          learningPathId,
+          userId: studentId,
+          actor,
+        }),
+      );
+    }
```
its-gabo: The event data is the same for both, so we can build it before-hand and then just publish start if isStarted and then return. By doing this we don't need the if isCompleted

apps/api/src/learning-path/services/learning-path.service.ts
```diff
@@ -771,19 +794,32 @@ export class LearningPathService {
...
+        newStudentIds.map((userId) =>
+          this.outboxPublisher.publish(
+            new EnrollLearningPathEvent({
+              learningPathId,
+              actor: currentUser,
+              userId,
+            }),
+            trx,
+          ),
+        ),
```
its-gabo: Let's add a bulk learning path enrollment event rather than publishing separate event for each enrollment (it can slow down the system). Let's do it similarly to `BulkUpdateCourseStatusEvent`

apps/api/src/scorm/scorm.service.ts
```diff
@@ -505,6 +512,20 @@ export class ScormService {
 
       await this.scormRepository.markPackageReady(jobData.packageId);
 
+      await this.outboxPublisher.publish(
+        new CreateScormEvent({
+          scormId: jobData.packageId,
+          actor: jobData.currentUser,
+          createdScorm: {
+            id: jobData.packageId,
+            entityType: SCORM_PACKAGE_ENTITY_TYPE.COURSE,
```
its-gabo: Let's derive the entity type from `jobData.action`

apps/api/src/scorm/scorm.service.ts
```diff
@@ -891,9 +938,23 @@ export class ScormService {
...
+    if (scoCompleted && !previouslyCompleted) {
+      await this.outboxPublisher.publish(
+        new CompleteScormEvent({
+          scormId: body.packageId,
+          actor: currentUser,
+          userId: currentUser.userId,
+        }),
+        this.db,
+      );
+    }
```
its-gabo: Let's publish the event when all SCOs are completed. For example the SCORM can have 15 SCOs and in this case we would publish the same event 15 times which is not really what we want.

apps/api/src/courses/course.service.ts
its-gabo: Also publishing the event only for single course delete. Let's handle bulk course deletion too. When adding the bulk one - then just publish one event as I described in one of other comments

its-gabo (changes requested):
Overally good, but few things need adjustement

its-gabo (changes requested):
The activity-log coverage is a solid, but several common flows still produce incomplete or misleading audit entries:
 
- Multi-SCO SCORM completion - completion is logged per lesson instead of once for the entire package
- Bulk SCORM deletion -  only the first SCORM package ID is retained, so the remaining deleted packages have no corresponding audit entry
- Learning-path snapshots - snapshots contain raw translation maps - use `localizationService.getLocalizedSqlField`
- Group-based enrollments - enrollments created through groups are not logged, leaving a gap in the learning-path enrollment history

its-gabo (approved):

PR #1823, opened 2026-07-29
chore: remove redundant category archiving functionality

its-gabo:
## Issue(s)

[#1817](https://github.com/Selleo/mentingo/issues/1817)

## Overview

Removes the redundant category archive workflow across the database schema, API contract, admin UI, generated client, and category test fixtures.

Archived categories are retained and become available after the migration drops the archive field. Category deletion remains the only removal workflow and continues to be blocked while a category is assigned to courses.

## Business Value

Category management is simpler and consistent across course configuration, administration, and global search. Administrators no longer need to maintain an archive state that does not provide a separate business workflow.

## Video

https://github.com/user-attachments/assets/2d740609-c9d9-419b-a51a-cadfecaab7a5


apps/api/src/category/category.service.ts
```diff
@@ -181,14 +178,7 @@ export class CategoryService {
...
 
-    const updateData: {
-      archived?: boolean;
-      title?: ReturnType<typeof setJsonbField>;
-    } = {};
-
-    if (updateCategoryBody.archived !== undefined) {
-      updateData.archived = updateCategoryBody.archived;
-    }
+    const updateData: { title?: ReturnType<typeof setJsonbField> } = {};
```
Japrolol: export type JsonbFieldUpdate = ReturnType<typeof setJsonbField>;

Japrolol (approved):

PR #1831, opened 2026-07-30
fix: prevent app crashes by safely processing each organization’s data access

Japrolol:
## Issue(s)
[1511](https://github.com/Selleo/mentingo/issues/1511)

## Overview
Replaced request-wide tenant transactions with tenant-only AsyncLocalStorage context. The tenant-aware Drizzle PgSession now opens a short transaction at terminal query execution, applies `SET LOCAL app.tenant_id`, reuses an active transaction for explicit service boundaries, and fails closed when tenant context is missing or conflicting.

Existing services and repositories keep using the `DB` token; no broad call-site migration or repository proxy is required. Added focused runner/session tests, an API-level tenant isolation test for duplicate emails, and updated the implementation/risks documentation.

Validation completed: API pre-commit typecheck, ESLint, Prettier, 70 API suites / 407 tests, and web tests with 271 passing tests.

## Business Value
Prevents request-wide idle-in-transaction connections from exhausting the application pool while retaining PostgreSQL RLS enforcement. Queries release connections promptly, concurrent requests are no longer limited by long-lived request transactions, and explicit multi-query transactions remain available where atomicity is required. The change also adds regression coverage for cross-tenant data leakage.


docs/implementation/tenant-rls-transaction-boundaries.md
```diff
@@ -0,0 +1,569 @@
+# Tenant RLS Transaction Boundaries
```
its-gabo: Please remove this file

its-gabo (approved):

PR #1878, opened 2026-08-13
feat: recover voice mentor and transcript streaming after reconnect

Japrolol:
## Issue(s)
[](https://github.com/Selleo/mentingo/issues/)

## Overview
This PR restores reliable live voice capture behavior in the web client by adding server-driven capture-mode resolution and reconnect/recovery handling for voice streaming sessions.

The changes are scoped to `apps/web/app/modules/Voice` and include:
- New shared audio streaming contracts in `apps/web/app/modules/Voice/audio-stream.types.ts` for capture mode, lifecycle events, reconnect payloads, and error codes.
- Protocol updates in `apps/web/app/modules/Voice/voiceSocketProtocol.ts` to support capture mode resolution and reconnect payload creation.
- Refactor in `apps/web/app/modules/Voice/audio-stream.ts` to track chunk sequence/state, wait for server start confirmation before mic capture starts, support both continuous and VAD-segmented capture flows, and recover from reconnect/chunk-sequence issues.
- Hook updates in:
  - `apps/web/app/modules/Voice/hooks/useVoiceMentor.ts`
  - `apps/web/app/modules/Voice/hooks/useTranscription.ts`
  to pass explicit capture mode and align frontend behavior with session mode.
- Follow-up fix (`fix: wait for resolved voice capture mode`) to prevent capture from starting until the session mode is resolved.

## Business Value
This makes voice mentor and transcription sessions more stable for learners and content users when network conditions are unstable, by reducing lost/recovered audio and preventing mode mismatch at session startup.


apps/web/app/modules/Voice/audio-stream.ts
```diff
@@ -182,15 +156,150 @@ export class RealtimePCMStreamerWorklet {
...
+      if (!this.isSessionActive) {
+        return;
+      }
+
+      if (this.captureMode === AUDIO_CAPTURE_MODE.CONTINUOUS) {
+        await this.startContinuousCapture();
+      } else {
+        await this.ensureMicVad();
+        await this.micVad?.start();
+      }
```
its-gabo, line 210: We could use guard clause pattern here

apps/web/app/modules/Voice/audio-stream.ts
```diff
@@ -304,6 +432,300 @@ export class RealtimePCMStreamerWorklet {
...
+      return null;
+    }
+    current = current[key];
+  }
+  return typeof current === "number" && Number.isInteger(current) && current >= 0 ? current : null;
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === "object" && value !== null;
+}
```
its-gabo, line 929: Can we move it to utils, because I believe that it's used in many places

its-gabo (approved):

Issue #1853, opened 2026-08-06
Fix missing metadata in activity logs


maciejhubert:
Take a look at LND tenant. There is metadata missing in some of them:

<img width="1391" height="219" alt="Image" src="https://github.com/user-attachments/assets/6467da45-21df-4495-baeb-5fdcf0f75b59" />

Also, I did not find which course that user enrolled to. If you have troubles in finding the user, please contact me.

PR #1882, opened 2026-08-13
chore: show additional activity details in the activity logs table

its-gabo:
## Issue(s)

- #1853

## Overview

Adds readable resource names to Activity Logs while preserving the existing resource-type column and resource ID.

- Returns a nullable `resourceName` in activity-log API responses.
- Prefers the name captured in event metadata so historical logs keep the event-time value.
- Resolves missing names from current tenant-scoped resources in batches grouped by resource type, avoiding one query per activity-log row.
- Uses the localization service for localized resource fields and supports users, courses, chapters, lessons, announcements, groups, categories, Q&A, news, articles, article sections, live trainings, learning paths, and SCORM packages.
- Adds a dedicated resource-name table column and displays the resource name and ID in expanded row details.
- Simplifies the Before, After, and Context presentation by removing their outer cards and outermost object braces.
- Parses JSON-encoded metadata values recursively so arrays and objects are displayed without escaped JSON strings.
- Adds translated labels and fallbacks for all supported UI locales.

## Business Value

Administrators can identify the resource affected by an activity without interpreting an internal UUID. Historical names remain useful after a resource is renamed or deleted, while metadata-free events such as enrollments can still display the current resource name when it exists. This makes audit investigations faster and easier to understand.

## Screenshots / Video
Screenshot 2026-08-13 at 19 48 35 – ze zmianami

### DIFF
--- a/apps/api/src/activity-logs/activity-log-resource-name.service.ts
+++ b/apps/api/src/activity-logs/activity-log-resource-name.service.ts
@@ -1,9 +1,14 @@
 import { Inject, Injectable } from "@nestjs/common";
-import { ACTIVITY_LOG_RESOURCE_TYPES, type ActivityLogResourceType } from "@repo/shared";
+import {
+  ACTIVITY_LOG_RESOURCE_TYPES,
+  SCORM_PACKAGE_ENTITY_TYPE,
+  type ActivityLogResourceType,
+} from "@repo/shared";
 import { eq, inArray, sql } from "drizzle-orm";
 
 import { DatabasePg, type UUIDType } from "src/common";
 import { LocalizationService } from "src/localization/localization.service";
+import { DB } from "src/storage/db/db.providers";
 import {
   announcements,
   articleSections,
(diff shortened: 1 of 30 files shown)


apps/api/src/activity-logs/activity-log-resource-name.service.ts
```diff
@@ -0,0 +1,278 @@
...
+  users,
+} from "src/storage/schema";
+
+import type { ActivityLogResponse } from "./activity-logs.types";
+
+type ResourceReference = Pick<ActivityLogResponse, "resourceId" | "resourceType">;
+type NamedResource = { id: string; name: string };
+
+@Injectable()
+export class ActivityLogResourceNameService {
```
Japrolol, line 36: Extract types to separate file and use symbol for db

apps/api/src/activity-logs/activity-log-resource-name.service.ts
```diff
@@ -0,0 +1,278 @@
...
+      .select({
+        id: scormPackages.id,
+        entityId: scormPackages.entityId,
+        entityType: scormPackages.entityType,
+      })
+      .from(scormPackages)
+      .where(inArray(scormPackages.id, ids));
+
+    const courseIds = packages
+      .filter(({ entityType }) => entityType === "course")
```
Japrolol: we have entity type enum

Japrolol (approved):

PR #1968, opened 2026-09-09
fix: ensure courses can be shared correctly and display the right author

its-gabo:
## Overview

Fixes course sharing failures when a target tenant already contains a matching category under a different locale.

Also ensures shared courses display the original source-course author instead of the local target-tenant user assigned as the technical course owner. The export snapshot now normalizes author metadata, fills missing metadata from the source author, and supports existing exported courses whose metadata was stored as a double-encoded JSONB string.

The master-course E2E coverage now verifies that:

- author metadata is copied to the exported course;
- the target tenant's `/api/course/all` response displays the source author;
- the exported course remains read-only.

## Business Value

Administrators can reliably share courses between tenants without exports failing because of localized category differences. Recipients also see the actual course author across course listings instead of an unrelated administrator from their own tenant, preserving correct attribution and avoiding confusion.


apps/api/src/courses/master-course.repository.ts
```diff
@@ -586,6 +605,16 @@ export class MasterCourseRepository {
     return existingCategory;
   }
 
+  async findCategoryByLocalizedTitle(title: string) {
+    const [existingCategory] = await this.db
+      .select(getTableColumns(categories))
```
Japrolol:
```suggestion
      .select()
```

Japrolol (approved):

PR #1987, opened 2026-09-22
fix: master course sharing after quiz refactor

its-gabo:
## Overview

Fix master-course synchronization failing with a duplicate display-order constraint after quiz choice options are reordered or replaced.

Remove stale option mappings, move existing options to temporary positions using Drizzle queries, and apply the new order within a transaction. Preserve surviving option IDs and save mappings after the option transaction commits.

## Business Value

Shared courses can receive updated quiz answers without synchronization failures, keeping recipient courses aligned with the source curriculum.


apps/api/src/courses/master-course.repository.ts
```diff
@@ -932,6 +932,33 @@ export class MasterCourseRepository {
...
+      .from(assessmentQuestionChoiceOptions)
+      .where(inArray(assessmentQuestionChoiceOptions.questionId, questionIds))
+      .orderBy(asc(assessmentQuestionChoiceOptions.id));
+
+    let temporaryOrder = options.reduce(
+      (minimum, option) => Math.min(minimum, option.displayOrder),
+      0,
+    );
+
+    for (const option of options) {
```
Japrolol, line 952: Not a fan of n queries, but let's let it be for now I guess

Japrolol (approved):

